PR #2, opened 2021-06-16
initial design doc.

raulk:


docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+area.
+
+*This `Mount` is provided out-of-box by the DAG store.*
+
+**Lotus Mount implementation**
+
+A Lotus `Mount` implementation would be instantiated with a sector ID and a
+bytes range within the sealed sector file (i.e. the deal segment).
+
+Loading the CAR consists of calling the worker HTTP API to fetch the unsealed
```
aarshkshah1992, line 165: So, the: 

```
- Check if we already have a local/remote unsealed copy.
- Get a Reader to the local Unsealed bytes or by performing a ranged http query over the remote Unsealed bytes if it already exists.
- Unseal the local or remote sector file if we don't have an unsealed copy.
- Get a Reader to the local or remote unsealed bytes.
```

is all taken care of by the Lotus Mount implementation, right ? I am confused by the _"A transformation function may be provided in the **future** as a `RegisterOpt`that conducts the unsealing"_.  

Are you suggesting that it is the caller's responsibility to perform the Unsealing if no unsealed copy exists and have the Mount implementation simply return the Unsealed copy or can we hide that detail in the Lotus mount implementation ?

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+- Given an indexed CARv2 file, it should be possible to instantiate a
+  `ReadBlockstore` on it in a self-sufficient manner.
+- Given a CARv1 or unindexed CARv2, it should be possible to instantiate a
+  `ReadBlockstore` with an index provided by us.
+- It should be possible to write an indexed CARv2 in streaming fashion. The
+  CARv2 library should do the right thing depending on the declared write
+  characteristics and the output characteristics. For example:
+   1. the writer may declare that blocks are provided in depth-first order and
+      without uniqueness guarantees, but may state they want the output to be in
+      depth-first order AND deduplicated. In this case, the CARv2 library must
```
aarshkshah1992, line 353: Just want to point out that this should be _"in an order  determined by the given IPLD selector"_ here as Graphsync could be using any IPLD selector as given by the Graphsync client. 

It just so happens that the default selector we use in Filecoin Markets is the "depth-first" selector if the user dosen't mandate the use of a specific selector.

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+```go
+dagst.RegisterShard(key []byte, mount Mount, opts ...RegisterOpts) error
+```
+
+1. This method takes a shard key and a `Mount`.
+2. It initializes the shard in `Unavailable` state.
+3. Calls `mount.Info()` to determine if the mount is of local or remote type.
+4. Calls `mount.Stat()` to determine if the mount target exists. If not, it
+   errors the registration.
+5. If remote, it fetches the remote resource into the scrap area.
```
aarshkshah1992, line 99: IMO, users should have the option to do this lazily. For a use case like Lotus, we don't want to fetch a remote Unsealed sector file into the scrap area at Shard Registration time because the registration would be done when a storage deal is completed but the CAR file should be copied to the local scrap area only when a retrieval is requested.

But then, we do need to fetch the unsealed sector file at shard registration to Index it when a storage deal is done. In that case, I think the user call flow for a Lotus storage deal  would look like:

1. `dagst.RegisterShard`(fetch, Index and mark shard as available) followed by a:
2. `dagst. ReleaseShard` (release local scrap copy, mark shard as unavailable) .

Then, during  retrieval, the DAG access by the user would result into the DAG Store fetching the unsealed CAR file locally (scrap area copy).

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+it in a local scrap area. Currently, this may lead to actual unsealing on the
+Lotus worker cluster through the current PoRep (slow) if the piece is sealed.
+
+With a future PoRep (cheap, snappy) PoRep, sealing can be performed _just_ for
+the blocks that are effectively accessed, potentially during IPLD block access
+time. A transformation function may be provided in the future as a `RegisterOpt`
+that conducts the unsealing.
+
+A prerequisite to enable unsealing-on-demand possible is PoRep and index
+symmetry, i.e. the byte positions of blocks in the unsealed CAR must be
```
aarshkshah1992, line 176: in the sealed* CAR.

aarshkshah1992, line 176:
```suggestion
symmetry, i.e. the byte positions of blocks in the sealed CAR must be
```

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+This versatility is provided by an abstraction called `mount.Mount`, a pluggable
+component which encapsulates operations/logic to:
+
+1. `Fetch() (io.ReadCloser, error)` Load a CAR from its origin.
+2. `Info() mount.Info` Provides info about the mount, e.g. whether it's local or
+   remote. This is used to determine whether the fetched CAR needs to be copied
+   to a scrap area.
+3. `Stat() (mount.Stat, error)` Equivalent to a filesystem stat, provides
+   information about the target of the mount: whether it exists, size, etc.
+
```
aarshkshah1992, line 147: I think we should add the `Marshal` & `Unmarshal` API to the `mount.Mount` interface to make it easy for the DAGStore to implement Mount persistence.

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+Options to explore here include LMDB, BoltDB, or others. Here's what the
+persisted shard entry could look like:
+
+```go
+type PersistedShard struct {
+   Key   []byte
+   // Mount is a URL representation of the Mount, e.g.
+   //   file://path/to/file?opt=1&opt=2
+   //   lotus://sectorID?offset=1&length=2
+   Mount string
```
aarshkshah1992, line 193: 1. To make it easy to persist Lotus mount implementations,  we could build out an endpoint in Markets that takes care of the magic elaborated [above](https://github.com/filecoin-project/dagstore/pull/2/files#r655842956). We can then persist this endpoint.

2. I think a better idea is to leave the details of Mount persistence to callers as mentioned in [this comment](https://github.com/filecoin-project/dagstore/pull/2/files#r655875242) for future extensibility.

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+will be elaborated on in the future.
+
+```go
+type DAGAccessor interface {
+   Shard(key shard.Key) ShardAccessor
+}
+
+type ShardAccessor interface {
+   Blockstore() (ReadBlockstore, error)
+}
```
aarshkshah1992, line 328: We can add a `Close` here that calls `ReleaseShard` under the hood. It is the user's responsibility to call `Close` when it's done with the dag access.

docs/design.md
aarshkshah1992: - Also, important to note that as discussed with @willscot, the CARV2 writer wont care about the selector/block order in the first iteration. 

- The CARv2 writer assumes that the caller knows what it's doing (i.e. it's writing blocks based on an order determined by a valid IPLD selector and that the CARv2 file will form a coherent DAG in the end) and does NOT enforce any restrictions on the order of the writes.

- So really, in the first  immediate iteration, the writer will declare that there are no uniqueness guarantees and that it expects the output to be de-duplicated but will make no declarations about the selector /block order as the CARv2 writer does NOT care about it.

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+   the full-shard index with no offset indication.
+   
+   These are calculated externally (e.g. semantic indexing service) and supplied
+   to the DAG store for storage and safekeeping.
+
+   A shard can have infinite number of semantic indices associated with it. Each
+   semantic index is identifid and stamped with its generation data (rule and
+   version).
+
+   We acknowledge that semantic indices perhaps don't belong in the DAG store
```
aarshkshah1992, line 245: +1

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+Upon starting, the DAG store will load the catalogue from disk and will
+reinstantiate the shard catalogue, the mounts, and the shard states.
+
+### Scrap area
+
+When dealing with remote mounts (e.g. Filecoin storage cluster), the DAG store
+will need to copy the remote CAR into local storage to be able to serve DAG
+access queries. These copies are called _transient copies_.
+
+Readers access shards from the storage layer by calling
```
aarshkshah1992, line 209: Just want to note that the `Reader` here is the _"DAG access layer"_ described below and the not the end-user as we do not want to expose the low level scrap area API to our end-users.

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+content.
+
+It is possible to mount shards with CARs accessible through the local
+filesystem, detachable mounts, NFS mounts, distributed filesystems like
+Ceph/GlusterFS, HTTP, FTP, etc.
+
+This versatility is provided by an abstraction called `mount.Mount`, a pluggable
+component which encapsulates operations/logic to:
+
+1. `Fetch() (io.ReadCloser, error)` Load a CAR from its origin.
```
aarshkshah1992, line 141: We should add a `FetchIndex` API here. If `FetchIndex` dosen't get us anything/errors out, we can build the Index based on the CAR we get from `Fetch` during Shard registration.

_Rationale:_

For an implementation like Lotus, even though Graphsyc writes out a CARv2 file, the Lotus mount `Fetch` implementation will only fetch the CARv1 payload as that's what's sealed into a sector. So,  during shard registration, what the DAG Store will see is a CARv1 payload and re-build the Index even though we had it in the original CARv2 thus doing wasteful work.

docs/design.md
aarshkshah1992: Note: We can build out an endpoint in Markets that takes care of all the above magic and the Lotus mount implementation ONLY needs to worry about calling that endpoint.

docs/design.md
```diff
@@ -0,0 +1,371 @@
...
+The choice of version and characteristics affects how the shard index is
+populated:
+
+1. CARv1: the index is calculated upon shard activation.
+2. Indexless CARv2: the index is calculated upon shard activation.
+3. Indexed CARv2: the inline index is adopted as-is as the shard index.
+
+### Shard states
+
+1. **Available:** the system knows about this shard and is capable of serving
```
aarshkshah1992, line 74: The shard state is never exposed to the DAG store client, right ? This state is only used for internal bookkeeping by the DAG Store. 

So, for example, when the user calls the DAG access layer for retrieval by a shard key, the DAG store will look up the shard state to determine if it's Available. If it's not available, the dag access layer can ask the scrap area to fetch the  CAR file (by using the `Mount` to Stat/Info/Fetch etc.) and then mark the shard state as `Available`.

Basically, there will be no `IsAvailable` or `MarkAvailable` etc APIs as the details of the shard state and how that state is changed are hidden by the DAG Store and only used for internal bookkeeping.

dirkmc:
Aarsh has already covered the specific comments I was going to make.

In general it would be nice if
- The DAG store interface allows the caller to pre-initialize things (eg loading an unsealed copy) but also has a mechanism to lazy load if necessary.
  - It should be possible to supply a flag that can enable / disable lazy loading.
  - If lazy loading is disabled and something is not already loaded, the method should return a specific error so that the calling code can do the load itself.
- There are methods for indices / stores / scrap data etc to
  - get the location on disk
  - get the disk usage
  - remove
 
In lotus miners typically do a lot of manual cleanup, eg deleting unsealed files, caches etc, so it would be nice if the interfaces provide recoverability from these situations.

docs/design.md
raulk: Exactly 👍  We should provide status accessors to query the state of a shard from the outside, but only on an informational level.

docs/design.md
raulk: Adding a `FetchIndex()` method here would break encapsulation as the `Mount` would now need to know how to seek to an index (which the shard management layer already knows how to process CARv1, CARv2-index, CARv2+index).

Also, all `Mount` implementations would need to duplicate this logic.

Instead, I think of the `Mount` as a low-level primitive that allows us to access remote byte streams.

I think it might be better to add a `io.Seeker` interface here that allows the shard management layer to:

1. Consume the pragma + header.
2. Decide if it wants only the index. If so, seek to the right position and read from there.

docs/design.md
raulk: Yup, already taken into account in the skeleton, but in a slightly different manner (using a registry).

docs/design.md
raulk: Yes, this is accurate 👍 

Just to note that Mounts do not manage transient copies (i.e. what you call "CAR not available locally"). I think we're both on the same page here, but wanted to confirm.

docs/design.md
raulk: Thanks!

docs/design.md
raulk: @willscott Just to be clear, the first iteration _will_ deduplicate blocks but will not reorder them (as we can make the hard guarantee that blocks are delivered in the order they are to be written), right?

docs/design.md
raulk: We don't need to persist behaviours, we need to persist state and have a mechanism to bind state to their implementations. Luckily, `Mounts` are conceived as stateless entities, so we only really need to persist the _configuration_ of a mount, which is done via URLs.

I've fleshed out more about this in the design doc now.

docs/design.md
raulk: That sounds good to me, adding.

raulk:
> It should be possible to supply a flag that can enable / disable lazy loading.

Initializing/filling a shard from its Mount (i.e. immediately downloading a transient copy if one is not available) can be a `RegisterOpt`.

> There are methods for indices / stores / scrap data etc to

Possibly, we'll have to see as we go!

docs/design.md
willscott: The current carbon will write a block everytime you call `Put`, so won't automatically dedup for you.
Does it make more sense for you as the caller to do the `Has` check to skip duplicates, or do we want that logic in the library?

@masih - I haven't looked closely at the method for opening the carv2 as a blockstore. Do we anticipate there'll be a reasonably ergonomic place to pass in an option for deduplicating puts when we do that?

I don't think this is hard, it's just a matter of where we do it.

docs/design.md
raulk: 👍  Yes, this sounds correct.

dirkmc (approved):

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+type ShardAccessor interface {
+   // Close() the accessor, signalling to the DAG store that we're done here and
+   // possible transient copies can be reclaimed once the shard is released.
+   io.Closer
+
+   Blockstore() (ReadBlockstore, error)
+}
+```
+
+## Requirements for CARv2
```
masih, line 383: Would selective car writer API be also required?

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+
+- Index needs to be detachable.
```
masih, line 385: Index readable as a ReaderAt via [`Reader`](https://github.com/ipld/go-car/blob/wip/v2/v2/reader.go#L61).

Do we need similar functionality in the CAR v2 blockstore API?

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+- Index offsets need to be relative to the CARv1, and not absolute in the
+  physical CARv2 file.
```
masih, line 387: ✅

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+- Index needs to be iterable.
```
masih, line 388: Index is a map; is there a specific API you have in mind for this?

docs/design.md
raulk: > Do we need similar functionality in the CAR v2 blockstore API?

Mind elaborating what this would look like?

docs/design.md
raulk: Mind defining "selective car writer"?

docs/design.md
raulk: We were hoping that the index would not have to be loaded into memory (which a map is). Something like this would work: https://github.com/filecoin-project/dagstore/pull/2/files#diff-b0929dacd4f1fc192cd308c1a9db2e85154ea742ccec315518e659d7f6a3efc7R347

docs/design.md
masih: A CAR v2 blockstore can be made configurable to either include the index in the final CAR v2 or not, and/or provide the dump of it as an `io.Reader`.

raulk: Gotcha, you mean on the write-side of the Blockstore. It wouldn't hurt and we might end up using it.

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+#### Shard destruction
+
+To destroy a shard, call:
+
+```go
+dagst.DestroyShard(key []byte) (destroyed bool, err error)
+```
+
+This erases transient copies from the scrap area, and it removes the shard/mount
+from the shard persistence store.
```
raulk, line 121:
```suggestion
This erases transient copies from the scrap area, and it removes the shard/mount
from the shard persistence store. It errors if the shard is locked/acquired.
```

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+
+```
+scheme://host[:port]?key1=value1&key2=value2[&...]
+```
+
+- `scheme` is a unique identifier for the mount type, e.g. `lotus://`,
+  `file://`, `http://`.
+- the rest is scheme-dependent.
+  - `host[:port]` is usually the main component of the mount, e.g. a file path,
+    a sector ID, etc.
```
raulk, line 205:
```suggestion
    a sector ID + deal ID, etc.
```

docs/design.md
masih: yes sorry the write-side of the blockstore I should have been clearer 😊 Thanks @raulk

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+   
+2. **Semantic indices.** Manifests of externally-relevant CIDs contained within
+   a shard, i.e. `[]CID` with no offset indication. In other words, subsets of
+   the full-shard index with no offset indication.
+   
+   These are calculated externally (e.g. semantic indexing service) and supplied
+   to the DAG store for storage and safekeeping.
+
+   A shard can have infinite number of semantic indices associated with it. Each
+   semantic index is identifid and stamped with its generation data (rule and
```
raulk, line 271:
```suggestion
   semantic index is identified and stamped with its generation data (rule and
```

docs/design.md
```diff
@@ -0,0 +1,423 @@
...
+
+In the future, this subcomponent will also serve the top-level cross-shard
+index.
+
+### Persistence and reconciliation
+
+Indices will be persisted in a filesystem directory, referenced by their shard
+key. This makes for easy identification.
+
+Although not expected, it is possible that over time the index repo and the shard management layer state will diverge (entropy):
```
raulk, line 298:
```suggestion
Although not expected, it is possible that over time the index repo and the
shard management layer state will diverge (entropy):
```

docs/design.md
masih: Car v2 version of [this](https://github.com/ipld/go-car/blob/master/selectivecar.go).

docs/design.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+## Requirements for CARv2
+
+- Index needs to be detachable.
+- Index offsets need to be relative to the CARv1, and not absolute in the
+  physical CARv2 file.
+- Index needs to be iterable.
+- Given any CAR file, we should be able to determine its version and
+  characteristics (concretely, indexed or not indexed).
```
masih, line 391: See: https://github.com/ipld/go-car/pull/90

A small clarification: characteristics won't specify whether the CAR v2 is indexed or not; instead this is determined by whether the `IndexOffset` is zero in the CAR v2 header.

docs/design.md
```diff
@@ -0,0 +1,424 @@
...
+- Given a CARv1 file, it should be possible to generate a CARv2 index for it in
+  detached form.
+- Given a CARv2 file, we should be able to decompose it to the corresponding
+  CARv1 payload and the Carbs index. The CAR payload should be exposed with
```
masih, line 395: Provided by [`Reader.CarV1ReaderAt`](https://github.com/ipld/go-car/blob/wip/v2/v2/reader.go#L56)

docs/design.md
aarshkshah1992: @masih

Yes, a Selective Car API that  takes a (root Cid, selector, read-only random access Blockstore) as input and writes out a  CARv2 would be of great help. 

Basically, it should look and feel exactly like the CARv1 selective CAR APIs but should write out a v2 instead of a v1.

docs/design.md
aarshkshah1992: @masih

The Carbs Index API/impl is a good reference:

We also need the ability to Marshal/Unmarshal Indexes so we can persist them to disk/read them from disk if need be.

masih: Captured: https://github.com/ipld/go-car/issues/95

docs/design.md
masih: Another related question: would it be OK for the write-side of blockstore to ask for the full list of roots when blockstore is constructed, like [carbon does](https://github.com/willscott/carbon/blob/main/carbon.go#L33)?

aarshkshah1992: @masih Yes, we can give you the roots up-front.

docs/design.md
rvagg: Can you clarify "depth-first" please? This could mean that visited blocks are yielded before or after visiting child blocks. We need to be clear about this point if we're baking this in anywhere and not just relying on selectors to do this for us.

docs/design.md
```diff
@@ -0,0 +1,432 @@
...
+  `io.ReaderAt` and `io.Reader` abstractions.
+- Given an indexed CARv2 file, it should be possible to instantiate a
+  `ReadBlockstore` on it in a self-sufficient manner.
+- Given a CARv1 or unindexed CARv2, it should be possible to instantiate a
+  `ReadBlockstore` with an index provided by us.
+- It should be possible to write an indexed CARv2 in streaming fashion. The
```
rvagg, line 401: If this is strictly necessary then you'll have to know the CARv1 contents size before writing the header because that's included there along with the index offset. SelectiveCar I think had a method to "prepare" so you could, in theory, get this information with a first-pass prior to actual writing. Where possible, however, rewinding and updating the header would be much more efficient.

aarshkshah1992 (approved):

docs/design.md
raulk: The way we expect this to happen is that the index is written in a separate temp file and then concatenated with the output file on finalisation. That enables the output file to grow on the payload portion (CARv1) unbounded, thus supporting the streaming scenario.  On finalisation we go back and update the header with the final offsets/sizes of everything.

willscott: noting that the carbon prototype will by default build up the index in memory - there's a hook where you can checkpoint it into a separate file as you need to, but because the index is getting insertion sorted on each new item, it isn't optimizing for recovery.

I think that's okay, but if you either want
(a) guarantees that on each append persisted to disk the index is also durable at the current contents of the partial car
or
(b) expect the index to not fit in memory
then there's a different write-ahead-log approach that should be used.

i think in reality this is fine and we just build up the index in memory during receive, and if we need to resume we re-build the partial index from the partial car.

masih: Is it OK to postpone this in the first pass or index in memory is a deal breaker for stream writes from get-go?

PR #3, opened 2021-06-16
[WIP] Carv2 requirements

aarshkshah1992:
CARv2 requirements for the Lotus native CAR store project.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -195,6 +195,14 @@ Local paths are currently prioritised over other forms. It is presumed
 - Index offsets need to be relative to the CARv1, and not absolute in the
   physical CARv2 file.
 - Index needs to be iterable.
+- Given a CARv2 file, we should be able to decompose it to the corresponding Carv1 payload and the Index.
+- Given a Carv1 file, we need to get it's Index.
+- Given a CARv2 file, we want a ReadOnlyBlockstore on top of it.  
+- Given a CARv1 file and it's Index, we want a ReadOnlyBlockstore wrapper on top if it.
+- We want a ReadWrite blockstore interface that can give us a CarV2 file when we commit. This is important for Graphsync.  
+    Need to flesh out whether we can write arbitary IPLD blocks or whether only coherent DAGs based on (root, selector) combo are allowed.
+- We should be able to read the leading N bytes of a CAR file and determine if it's a CARv1 or a Carv2.
+- TODO: Give a thought to the CommP calculation on both sides in the storage deal -> is probably covered by the requirements above.
 
 ## Brainstorm
 


docs/design.md
```diff
@@ -195,6 +195,14 @@ Local paths are currently prioritised over other forms. It is presumed
 - Index offsets need to be relative to the CARv1, and not absolute in the
   physical CARv2 file.
 - Index needs to be iterable.
+- Given a CARv2 file, we should be able to decompose it to the corresponding Carv1 payload and the Index.
+- Given a Carv1 file, we need to get it's Index.
+- Given a CARv2 file, we want a ReadOnlyBlockstore on top of it.  
+- Given a CARv1 file and it's Index, we want a ReadOnlyBlockstore wrapper on top if it.
+- We want a ReadWrite blockstore interface that can give us a CarV2 file when we commit. This is important for Graphsync.  
+    Need to flesh out whether we can write arbitary IPLD blocks or whether only coherent DAGs based on (root, selector) combo are allowed.
+- We should be able to read the leading N bytes of a CAR file and determine if it's a CARv1 or a Carv2.
```
willscott, line 204: Do you care about version, or do you care about if it has an index already?

docs/design.md
```diff
@@ -195,6 +195,14 @@ Local paths are currently prioritised over other forms. It is presumed
 - Index offsets need to be relative to the CARv1, and not absolute in the
   physical CARv2 file.
 - Index needs to be iterable.
+- Given a CARv2 file, we should be able to decompose it to the corresponding Carv1 payload and the Index.
```
willscott, line 198: what form of decomposition do you need? is a `ReaderAt` providing streaming or random access to the Carv1 payload sufficient, or do you want to realize the carv1 payload on disk?

raulk: I think a `ReadSeekCloser` + `ReaderAt` would be most flexible.

docs/design.md
raulk: If it's a CARv2 you will need to determine whether it has an index or not. Possibly a:

```go
car.Info(io.Reader) (version, Characteristics)
```

Would make sense as a function, or the equivalent on a CAR object if we want it as a method?

willscott: we might do it as a method on a parsed `Header`, so that you only need to read that once, since you'll be parsing the header to get the `Info` anyway.

so concretely [`ReadHeader(br *bufio.Reader) (*CarHeader, error)`](https://github.com/ipld/go-car/blob/master/car.go#L72) would have a `Version()` that tells you if it's either v1 or v2 that's been read, and would have a `HasIndex() bool` or similar.

raulk:
Absorbed into #2.

PR #4, opened 2021-06-22
skeleton: mounts.

raulk:
Based on the design doc.

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -1,0 +1,73 @@
+package mount
+
+import (
+	"io"
+	"net/url"
+)
+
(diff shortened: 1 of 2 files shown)


dirkmc:
Overall LGTM 👍 

I'm wondering if we need the registry - do we anticipate needing to load mount types dynamically at runtime?

dirkmc:
Can we just use a hardcoded list of mount implementations instead of having the registry?

raulk:
Unfortunately, no, because the Lotus mount is externally provided (by Lotus) -- despite being incubated in this repo, it does not belong here, as it's Lotus specific.

PR #10, opened 2021-06-22
skeleton: index repo.

raulk:


index/repo.go
```diff
@@ -0,0 +1,50 @@
...
+type FullIndexRepo interface {
+	// GetFullIndex returns the full index for the specified shard.
+	GetFullIndex(key shard.Key) (FullIndex, error)
+
+	// AddFullIndex persists a full index for a shard.
+	AddFullIndex(key shard.Key, index FullIndex) error
+
+	// DropFullIndex drops the full index for the specified shard. If the error
+	// is nil, it returns whether an index was effectively dropped.
+	DropFullIndex(key shard.Key) (dropped bool, err error)
```
dirkmc, line 28: Question: in which scenario is dropped false?

index/manifest.go
```diff
@@ -0,0 +1,33 @@
...
+// ManifestKey identifies a manifest. It is a triple that can act as a composite
+// key, comprising the shard key and generation metadata.
+type ManifestKey struct {
+	Shard      shard.Key
+	GenRule    string
+	GenVersion uint
+}
+
+// Manifest are sets of CIDs with no offset indication.
+type Manifest interface {
```
dirkmc, line 17: We've captured the concept of the Manifest in the design docs.
I suggest we don't add interfaces / implement the Manifest until we need it in practice.

index/full.go
```diff
@@ -0,0 +1,39 @@
+package index
+
+import (
+	"errors"
+
+	"github.com/ipfs/go-cid"
+)
+
+// ErrNotFound is returned when an entry is not found in an index.
+var ErrNotFound = errors.New("entry not found in index")
```
dirkmc, line 10: Suggest we create a custom error struct:
```
type ErrNotFound struct {
    Message string
    c cid.Cid
}
```

index/full.go
```diff
@@ -0,0 +1,39 @@
...
+
+// TODO we may be able to replace this interface with once it's public, as long
+//  as it doesn't leak implementation details.
+//  https://github.com/ipld/go-car/blob/2a49cbd958b0c6443dd1e9b518814199bf50842d/v2/internal/index/index.go#L37
+type FullIndex interface {
+	// Offset returns the offset of a CID inside the data CAR of a shard.
+	//
+	// If the CID exists, the offset will be zero or positive.
+	// If the CID is not contained in the index, it will return an
+	// ErrCidNotFound error
```
dirkmc, line 20:
```suggestion
	// ErrNotFound error
```

index/repo.go
```diff
@@ -0,0 +1,50 @@
...
+	Exists bool
+	Size   uint64
+}
+
+type FullIndexRepo interface {
+	// GetFullIndex returns the full index for the specified shard.
+	GetFullIndex(key shard.Key) (FullIndex, error)
+
+	// AddFullIndex persists a full index for a shard.
+	AddFullIndex(key shard.Key, index FullIndex) error
```
dirkmc, line 24: I think we will need to add a `Write` method to `FullIndex` in order for `AddFullIndex` to be able to write the index to disk

index/repo.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Repo is the central index repository object that manages full indices and
+// manifests.
+type Repo struct {
+	FullIndexRepo
+	ManifestRepo
+}
+
+type Stat struct {
+	Exists bool
```
dirkmc, line 15: Question: When is Exists false?

index/repo.go
```diff
@@ -0,0 +1,50 @@
...
+	Size   uint64
+}
+
+type FullIndexRepo interface {
```
dirkmc, line 19: I suggest we include methods to
- get the count of indexes in the repo
- iterate over index keys
- get the size of the repo

index/manifest.go
raulk: I would prefer to draw the shape of the system entirely even if we don't implement them. We can comment out the embedding of the Manifest interface in the `index.Repo`.

index/full.go
raulk: How about we use error wrapping:

```
return fmt.Errorf("entry %s: %w", cid, ErrNotFound).
```

https://blog.golang.org/go1.13-errors

index/repo.go
raulk: I intended for this to be similar to the interface of a `DELETE` command in most databases, which tells you how many rows were affected. However, since there can only be one full index, it's a binary and not a number.

In terms of logical coherence, there shouldn't be any case where `DropFullIndex` returns false, as that would indicate that an index was lost. But it may happen if the user erased an index manually from disk! For those cases, we have planned a `ReconcileIndices()` method: https://github.com/filecoin-project/dagstore/blob/master/docs/design.md#persistence-and-reconciliation

index/repo.go
raulk: When the index exists on disk. This may be a little bit of YAGNI FWIW, but useful for testing. Feel free to panic in the implementation when `Stat` is called now. Then we can decide if we do need it or not.

index/repo.go
raulk: Agree, wanna push a change here?

index/repo.go
raulk: Makes sense. We do want to transition to using the CARv2 Index abstraction (import from CARv2), which I believe provides that: https://github.com/ipld/go-car/blob/2a49cbd958b0c6443dd1e9b518814199bf50842d/v2/internal/index/index.go#L37-L43

index/repo.go
raulk: These changes are pushed as part of #16.

raulk:
Looks like all comments are addressed, merging.

PR #11, opened 2021-06-23
Finish registry and tests

aarshkshah1992:
- Finishes the registry implementation and adds tests.
- Init go mod.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,8 @@
+module github.com/filecoin-project/dagstore
+
+go 1.16
+
+require (
+	github.com/stretchr/testify v1.7.0
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
+)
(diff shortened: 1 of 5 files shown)


dirkmc:
LGTM 👍 

One suggestion I have is to use the Factory pattern for the registry so that the dagstore doesn't need to know about how to make API calls when fetching. This would also allow us to avoid using reflection:
```golang
type MntFactory interface {
	Parse(url string) mount.Mount
}

type ExpMntAPI interface {
	MakeSomeAPICall() error
}

type ExpMntFactory struct {
	api ExpMntAPI
}

func NewExpMntFactory(api ExpMntAPI) *ExpMntFactory {
	return &ExpMntFactory{api: api}
}

func (f *ExpMntFactory) Parse(url string) mount.Mount {
	if url[0:6] == "expmnt" {
		return &ExpMnt{api: f.api}
	}
	return nil
}

type ExpMnt struct {
	api ExpMntAPI
}

func (e *ExpMnt) Fetch() {
	e.api.MakeSomeAPICall()
	// ...
}

func (r *Registry) Register(f MntFactory) {
	r.factories = append(f.factories, f)
}

func (r *Registry) Instantiate(url string) mount.Mount {
	for _, f := range r.factories {
		if mnt := f.Parse(url); mnt != nil {
			return mnt
		}
	}
	return nil
}

func Start() {
	api := // create API
	registry.Register(NewExpMntFactory(api))
}
```

aarshkshah1992:
Subsusmed by https://github.com/filecoin-project/dagstore/pull/12.

PR #12, opened 2021-06-23
Registry that uses a Mount Factory to instanitiate a Mount from the deterministic Mount URL for a Shard

aarshkshah1992:
Subsusmes https://github.com/filecoin-project/dagstore/pull/11.

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -1,6 +1,7 @@
 package mount
 
 import (
+	"context"
 	"io"
 	"net/url"
 )
@@ -49,7 +50,7 @@ const (
 type Mount interface {
 	// Fetch fetches the asset from its original location, returning a read only
 	// stream. This may be a remote or local stream.
-	Fetch() (io.ReadCloser, error)
+	Fetch(ctx context.Context) (io.ReadCloser, error)
 
 	// Info describes the Mount. This is a pure function.
 	Info() Info
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -45,7 +45,7 @@ const (
 //
 // Mounts must define a deterministic URL representation which will be used to:
 //
-//  a. deserialise the Mount from DAG persistence when resuming the system.
+//  a. deserialise the Mount from DAG persistence when resuming the system by using a pre-defined Mount factory mapped to the URL scheme.
 //  b. support adding mounts from configuration files.
 type Mount interface {
 	// Fetch fetches the asset from its original location, returning a read only
@@ -57,9 +57,6 @@ type Mount interface {
 
 	// Stat describes the remote resource.
 	Stat() (Stat, error)
-
-	// Parse initializes the mount from a URL. This is a pure function.
-	Parse(url *url.URL) error
 }
 
 // Info describes a mount.
@@ -71,4 +68,10 @@ type Info struct {
 type Stat struct {
 	Exists bool
 	Size   uint64
+}
+
+// MountFactory allows instantiation of a Mount from the deterministic URL representation defined by the Mount.
+type MountFactory interface {
+	// Parse initializes the mount from a URL.
+	Parse(u *url.URL) (Mount, error)
 }
--- a/mount/registry.go
+++ b/mount/registry.go
@@ -4,25 +4,46 @@ import (
 	"errors"
 	"net/url"
 	"reflect"
+	"sync"
+
+	"golang.org/x/xerrors"
 )
 
 // ErrUnrecognizedScheme is returned by Instantiate() when attempting to
 // initialize a Mount with an unrecognized URL scheme.
 var ErrUnrecognizedScheme = errors.New("unrecognized mount scheme")
 
+// ErrRegistrationWithNonPointerType is returned by Register() when attempting to
+// register a Mount with a dynamic non-pointer type.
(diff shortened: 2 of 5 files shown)


mount/registry.go
```diff
@@ -3,32 +3,55 @@ package mount
...
+	defer r.lk.RUnlock()
+
+	mft, ok := r.m[u.Scheme]
+	if !ok {
+		return nil, ErrUnrecognizedScheme
+	}
+
+	mt, err := mft.Parse(u)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to instantiate mount, factory.Parse, err=%s", err)
```
dirkmc, line 52:
```suggestion
		return nil, xerrors.Errorf("failed to instantiate mount with factory.Parse: %w", err)
```

dirkmc (approved):
LGTM 👍 

Just one nit

mount/mount.go
```diff
@@ -44,21 +45,18 @@ const (
...
 // Mounts must define a deterministic URL representation which will be used to:
 //
-//  a. deserialise the Mount from DAG persistence when resuming the system.
+//  a. deserialise the Mount from DAG persistence when resuming the system by using a pre-defined Mount factory mapped to the URL scheme.
 //  b. support adding mounts from configuration files.
 type Mount interface {
 	// Fetch fetches the asset from its original location, returning a read only
 	// stream. This may be a remote or local stream.
-	Fetch() (io.ReadCloser, error)
+	Fetch(ctx context.Context) (io.ReadCloser, error)
```
raulk, line 53: Good call.

mount/mount.go
```diff
@@ -71,3 +69,9 @@ type Stat struct {
 	Exists bool
 	Size   uint64
 }
+
+// MountFactory allows instantiation of a Mount from the deterministic URL representation defined by the Mount.
+type MountFactory interface {
+	// Parse initializes the mount from a URL.
+	Parse(u *url.URL) (Mount, error)
+}
```
raulk, line 77: I'm OK going this direction for now, but I think it introduces unnecessary complexity.

What I had in mind was using reflection and the type registry to instantiate an empty struct of whichever kind, and then calling `Parse` on it to populate it. This is a bit reminiscent to JSON unmarshalling in Go.

Simplifying just for the purposes of illustration:

```
var registry map[string]reflect.Type // registry, where the key is the scheme.
var url *url.URL // url we want to parse

typ, _ := registry[url.Scheme]
mnt := reflect.New(typ).Interface().(Mount)
mnt.Parse(url) // mnt consumes the URL to populate itself
```

dirkmc: The reason I suggested using the factory is so that you can inject an `api` object into the Mount.
This allows the calling code (in this case lotus) to abstract away the API call mechanism (HTTP, JSON-RPC etc) so that dagstore doesn't need to know about it.

raulk: Gotcha. I think the underlying requirement that you're identifying is that there needs to be a way to supply some environmental context to the mount. These cannot be persisted in the URL because they may change across restarts (e.g. the miner JSON-RPC endpoint changing). The mount factory is one way to encapsulate the context but I think we'll find it too inflexible. For example, what if the sealing cluster JSON-RPC API endpoint changes at runtime? I think we'll need something more powerful going forward. But happy to keep it like this for now.

mount/registry_test.go
```diff
@@ -0,0 +1,131 @@
...
+	r := Registry{
+		m: make(map[string]MountFactory),
+	}
+
+	// create & register mock mount factory 1
+	u := fmt.Sprintf("http://host1:123?size=%d", m1StatSize)
+	url, err := url.Parse(u)
+	require.NoError(t, err)
+	m1 := &MockMountFactory1{}
+	require.NoError(t, r.Register(url.Scheme, m1))
```
raulk, line 97: Better to hardcode the expected fixed value: Register("http", ...). Otherwise the test becomes rather circular.

raulk (approved):

mount/mount.go
raulk: https://github.com/filecoin-project/dagstore/issues/13

raulk:
Gonna go ahead and push over the finish line by applying the suggested changes.

mount/registry_test.go
raulk, line 97:
```suggestion
	require.NoError(t, r.Register("http", m1))
```

mount/registry_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// // register same scheme again -> fails
+	require.Error(t, r.Register(url.Scheme, m1))
```
raulk, line 99:
```suggestion
	require.Error(t, r.Register("http", m1))
```

mount/registry_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// create and register mock mount factory 2
+	url2 := fmt.Sprintf("ftp://host2:1234?size=%d", m2StatSize)
+	u2, err := url.Parse(url2)
+	require.NoError(t, err)
+	m2 := &MockMountFactory2{}
+	require.NoError(t, r.Register(u2.Scheme, m2))
```
raulk, line 106:
```suggestion
	require.NoError(t, r.Register("ftp", m2))
```

PR #16, opened 2021-06-23
Add full index repo implementations

dirkmc:
Implementations of a full index repo
- In-memory
- Filesystem

TODO:
- [x] Clarify whether we can assume that the entire index will be in-memory (as opposed to being on-disk)
  We can assume the entire index will be in-memory
- [ ] Add methods mentioned [here](https://github.com/filecoin-project/dagstore/pull/10/files/1759daf6970095df7acf537a1f0d6bda96e05ef6#diff-7441ed0d8da25f5e9300d892fbd0e5ee5ffd2036a0929f806f23d6adfd301b4d)
  I'm going to add these as a separate PR
- [x] Add more unit tests (eg repo version check)


impl/local/localindexrepo.go
```diff
@@ -0,0 +1,111 @@
+package local
```
raulk, line 1: Let's avoid impl packages; not idiomatic in Go. These should go in the `index` package directly. It's OK for it to contain the interface _and_ the implementations.

impl/test/indexrepotest.go
```diff
@@ -0,0 +1,57 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/filecoin-project/dagstore/index"
+	"github.com/filecoin-project/dagstore/shard"
+	"github.com/ipfs/go-cid"
+	"github.com/stretchr/testify/require"
+)
+
+func RunIndexRepoTest(t *testing.T, r index.FullIndexRepo) {
```
raulk, line 13: It's possible that we will have multiple implementations of the index repo going forward (such as when we move to run a distributed DAG store). Right now we have a "forced" one (memory repo), which we'll never use in production, but it's useful for tests. Instead of having test indirection, I would prefer to use subtests.

raulk (approved):
Besides the package structure, this looks good to me for a start. It will likely be modified heavily as we adopt the actual `car.Index` object here.

One thing I don't get is the usage of a factory pattern here too, looks like a little bit forced and it will probably be irrelevant once we start using CARv2?

dirkmc:
Whether we need the factory pattern depends on whether we can assume the index is entirely in-memory. For now I've merged the change to remove the factory, but we can always back that out. I also added a checkbox in the issue description to clarify this requirement.

raulk (approved):
Looks great for a starting point. It'll change as we flesh out other parts of the system, and we integrate with go-car v2.

index/repo_fs.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		closeErr := f.Close()
+		if err == nil {
+			err = closeErr
+		}
+	}()
```
raulk, line 76: Noticed this while integrating. We usually don't want to propagate this error, as the goal of the method may have actually been satisfied (we may be returning an index, and for whatever reason the OS may fail when closing the file handle).

PR #17, opened 2021-06-23
index repo: review of #6

raulk:
Commit log shows proposed changes.

- Simplified the package setup and file naming, as it was showing some repetitiveness and things were more spread out than they needed to be.
- introduce testify/suite
- subsume the index factory into the filesystem repo -- I don't think we need pluggability at this layer, as we will be relying on Carbs serde.


index/repo_fs.go
```diff
@@ -54,12 +52,30 @@ func NewLocalIndexRepo(factory FullIndexFactory, baseDir string) (*LocalIndexRep
...
+		if err == nil {
+			err = closeErr
+		}
+	}()
+
+	err = idx.Unmarshal(f)
+	if err != nil {
+		return nil, err
+	}
+	return idx, nil
```
dirkmc, line 75: I implemented it the way you have it here originally, but the problem is that this assumes that the entire index is in memory.
If instead we use a factory pattern, we can just give the index file path to the index implementation, and the implementation can read / seek over the file as needed without needing to load the entire index into memory. This also gives full control to the index over when to open and close the file handle.

dirkmc (changes requested):

dirkmc:
Merging as I think we will want most of this - if we decide we want to support indices that are not in memory we can back out that change.

PR #19, opened 2021-06-24
CAR implementation of FullIndex

dirkmc:
Adds a CAR implementation of `FullIndex`

Questions:
- Should the CAR implementation live within dagstore or should it be injected?
- Should we provide a registry / factory concept (so that we don't need to hardcode the CAR implementation as the index
- See TODOs and panics in code for things that need to be added to go-car


index/full_car.go
```diff
@@ -0,0 +1,57 @@
+package index
```
raulk, line 1: As I implement the DAG store, I realise that the FullIndex abstraction is actually unnecessary at this stage. In fact, it complicates things because when we obtain indices from the index repo to join with the data to create a read-only blockstore, and we do this through the carv2 library, which of course expects a carv2 index.

Conclusion: The index repo can custody CARv2 indices directly.

raulk (approved):
Will merge and then change in place. Thanks for this exploration!

PR #20, opened 2021-07-01
initial implementation of the dagstore.

raulk:
🐉🐉🐉🐉🐉🐉  BEWARE OF DRAGONS!

---

This PR contributes an initial implementation of the DAG store.

## Concurrency model

The DAG store operates on an event loop to perform mutations to the shard state. This is important as it allows shard operations to be non-blocking (although we can provide a blocking stub on top of this async plumbing). This means that the user does not need to worry about for a shard registration to have returned before they can go and ask to acquire that shard. I explored numerous alternatives and discarded them, such as:

* using locks all over the place; prone to concurrency bugs and a nightmare to debug races.
* using per-shard goroutines; this is wasteful as registering 1MM shards would immediately explode onto 1MM goroutines (which is quite feasible and expected for Filecoin miners, 1 deal = 1 shard). Furthermore, the majority of those goroutines would be idle most of the time.

## Finite state machine

Shards to go through a finite state machine. It would be worth to formalise this FSM.

## Public API

The public API is quite simple:

```go
RegisterShard(key shard.Key, mnt mount.Mount, out chan ShardResult, opts RegisterOpts) error
AcquireShard(key shard.Key, out chan ShardResult, _ AcquireOpts) error
DestroyShard(key shard.Key, out chan ShardResult, _ DestroyOpts) error
```

The caller supplies a channel where they want the result returned, when available. This is much better and flexibile than having the DAG store return a channel, since it enables the caller to send all results to a single channel with a single goroutine servicing it.

The caller can also pass a per-call channel if they so desire. Moreover, we can easily create sugar `*Sync()` method counterparts that do this behind the scenes, giving the appearance of a synchronous interface.

## Mount upgrader

This PR contributes a "mount upgrader". This is in turn a Mount that acts as a locally cached copy of a remote mount (e.g. Lotus, HTTP, FTP, NFS, etc.)

## Indices

I was able to remove the `FullIndex` abstraction, since the CAR library already provides an `index.Index` interface that we can use directly.

## TODO

This is an initial, unstable, buggy implementation to get the ball rolling. There are many features missing here, such as:

- Transient file tracking and cleanup; refcounting in the `mount.Upgrader`. Aka "scrap area".
    - #26
- mmap transients.
    - #28
- Proper failure management (and recovery).
    - #33
- ~Proper backpressure management (backlogged task channel).~
    - No longer needed, as internal and external channels have been split out in 258d74d.
- Correct shard destruction.
    - #31
- Shard release (via ShardAccessor#Close).
    - #27
- Shard state persistence and recovery upon restart.
    - #32
- Repopulate transients on restart from the scratch space.
    - #24
- Transient vs. mount integrity -- how do we recover from a corrupted transient?
    - #25
- Lots of unit tests.
- CI.
- Synchronous sugar on top of async API.


dagstore.go
```diff
@@ -0,0 +1,602 @@
...
+				err = fmt.Errorf("failed to add index for shard: %w", err)
+				_ = d.queueTask(&Task{Op: OpShardFail, Shard: tsk.Shard, Error: err}, true)
+				break
+			}
+			_ = d.queueTask(&Task{Op: OpShardMakeAvailable, Shard: tsk.Shard}, true)
+
+		case OpShardMakeAvailable:
+			if s.wRegister != nil {
+				res := ShardResult{Key: s.key}
+				go func(ch chan ShardResult) { ch <- res }(s.wRegister)
```
iand, line 417: Unbounded growth in goroutines here if client supplies a nil channel to RegisterShard or is slower than the dagstore at reading from the supplied channel

iand: An alternate API design that avoids the need for the dagstore to manage the lifetimes of these goroutines.

Current client pattern looks like this:

```
results := make(chan ShardResult, 100)

go func() {
	select {
		case res := <-results:
			// do something with the result
	}
}()

for _, key := range shards {
	dagstore.RegisterShard(key, mnt, results, opts)
}
```

Client MUST service the results channel otherwise the dagstore will continue to spawn goroutines. Cannot pass nil channel. Closing the dagstore will leave these goroutines orphaned.

Alternate design is to return a ticket which the client may optionally use to enquire the status of the shard operation:

```
RegisterShard(key shard.Key, mnt mount.Mount,opts RegisterOpts) (*ShardTicket, error)

```

A ticket is a concrete type:

```
type ShardTicket struct {
	mu       sync.Mutex     
	done     chan struct{}
	result   *ShardResult
}

// Done returns a channel that is closed when the shard operation is complete
func (s *ShardTicket) Done() <-chan struct{} {
	return s.done
}

// Result returns the result of the shard operation or nil if the operation has not completed
func (s *ShardTicket) Result() *ShardResult {
	s.mu.Lock()
	res := s.result
	s.mu.Unlock()
	return res
}

// complete marks the operation as complete
func (s *ShardTicket) complete(result *ShardResult) {
	s.mu.Lock()
	s.result = result
	s.mu.Unlock()
	close(s.done)
}
```

The dagstore simply calls `complete` on the ticket it created when the operation has been completed.

The client can make the dagstore call synchronous:

```
ticket, _ := dagstore.RegisterShard(key, mnt, results, opts)
select {
	case <-ticket.Done():
}
res := ticket.Result()
```

Or asynch, but here the number of goroutines is under the control of the client and can cancelled with to client context:

```
results := make(chan ShardResult, 100)

go func() {
	select {
		case res := <-results:
			// do something with the result
	}
}()


for _, key := range shards {
	ticket, _ := dagstore.RegisterShard(key, mnt, results, opts)
	go func() {
		select {
			case <-ticket.Done():
				results <- ticket.Result()
			case <-ctx.Done():
				return
		}
	}()
}
```

iand:
I suggest that the public API accepts contexts to facilitate request tracing, metrics and logging

dagstore.go
```diff
@@ -0,0 +1,602 @@
...
+
+func (d *DAGStore) queueTask(tsk *Task, overflow bool) error {
+	select {
+	case <-d.ctx.Done():
+		return fmt.Errorf("dag store closed")
+	case d.taskCh <- tsk:
+		return nil
+	default:
+	}
+	if overflow {
```
dirkmc, line 528: I think it would simplify the implementation to use a single queue instead of having a separate taskCh channel and overflow queue

mount/upgrader.go
```diff
@@ -0,0 +1,160 @@
...
+
+// Upgrade constructs a new Upgrader for the underlying Mount.
+func Upgrade(underlying Mount, initial string) (*Upgrader, error) {
+	ret := &Upgrader{underlying: underlying}
+
+	info := underlying.Info()
+	if !info.AccessSequential {
+		return nil, fmt.Errorf("underlying mount must support sequential access")
+	}
+	if info.AccessSequential && info.AccessSeek && info.AccessRandom {
```
dirkmc, line 39:
```suggestion
	if info.AccessSeek && info.AccessRandom {
```

mount/upgrader.go
```diff
@@ -0,0 +1,160 @@
...
+	defer file.Close()
+
+	u.transient = file.Name()
+
+	// sanity check on underlying mount.
+	if stat, err := u.underlying.Stat(ctx); err != nil {
+		return fmt.Errorf("underlying mount stat returned error: %w", err)
+	} else if !stat.Exists {
+		return fmt.Errorf("underlying mount no longer exists")
+	}
```
dirkmc, line 118: Can we rely on Fetch to return these errors if the file doesn't exist?

mount/upgrader.go
```diff
@@ -0,0 +1,160 @@
...
+			return os.Open(u.transient)
+		}
+	}
+
+	// transient appears to be dead, refetch.
+	if err := u.refetch(ctx); err != nil {
+		return nil, err
+	}
+
+	return os.Open(u.transient)
```
dirkmc, line 74: This feels a little racy - can we return a file handle from `u.refetch()`?

dagstore.go
raulk: @iand The same API usage can be achieved with the current model by passing in a per-request channel. As a user:

```go
ch := make(chan ShardResult, 1)
_ = dagstore.RegisterShard(key, mnt, ch, opts)
<-ch
```

With this model, the client is truly in control of how they want to consume results.

I deliberately don't want the DAG store to return a channel because that's the most inflexible model. If you register many shards at once, you will need to monitor each channel with a separate goroutine (unless you use reflection to select of many channels -- yuck).

raulk: BTW -- this PR does not guarantee order-safety, which I do want. It's likely that I end up implementing a notifier goroutine on the DAG side whose job is to publish results back to users. So if the user blocks, it won't block the DAG store's event loop, just the result stream.

dagstore.go
raulk: I have a better model! This can be simplified by using two chans:

- internal, with buffer 1, used for internal tasks.
- external, with buffer N, used for external tasks.

dagstore.go
iand: The code you show is identical to the synchronous example I gave except you require the client to know to use a correctly sized channel. 

The main issue I see is the unbounded growth of goroutines in a long running server easily caused by the client holding the API wrong (unbuffered / slow draining / nil channel). The dagstore has no way of detecting when these should be terminated, but the client does.

raulk: > The code you show is identical to the synchronous example I gave except you require the client to know to use a correctly sized channel.

Client can use any buffer size they want. To keep the system healthy, they should be consuming, but the system won't stop making progress if they fail to do so.

> The main issue I see is the unbounded growth of goroutines in a long running server easily caused by the client holding the API wrong (unbuffered / slow draining / nil channel).

I agree, and as I said above, I plan to move response dispatch to its own, single-threaded goroutine, instead of spinning up ad-hoc goroutines. Doing so will also preserve ordering.

> The dagstore has no way of detecting when these should be terminated, but the client does.

What introduces that ability in your example is the client passing in a `context.Context`, not the DAG store creating and owning a channel. I agree this might be something worth doing, also for the purposes of tracing/metrics, as you also pointed out.

dagstore.go
raulk: Solved here: https://github.com/filecoin-project/dagstore/pull/22

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+				_ = d.queueTask(&Task{Op: OpShardFail, Shard: tsk.Shard, Error: err}, d.internalCh)
+				break
+			}
+			// queue a fetch.
+			_ = d.queueTask(&Task{Op: OpShardFetch, Shard: tsk.Shard}, d.internalCh)
+
+		case OpShardFetch:
+			s.state = ShardStateFetching
+
+			go func(ctx context.Context, upgrader *mount.Upgrader) {
```
aarshkshah1992, line 351: Should use the waitGroup to refcount all go-routines we spawn.

dagstore.go
aarshkshah1992: > I agree, and as I said above, I plan to move response dispatch to its own, single-threaded goroutine, instead of spinning up ad-hoc goroutines. Doing so will also preserve ordering.

We will have to introduce some sort of backpressure at the point the event loop hands off the result to the result dispatcher go-routine which will block the event loop or start dropping results to the floor.

mount/upgrader.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (u *Upgrader) Close() error {
+	panic("implement me")
+}
+
+func (u *Upgrader) refetch(ctx context.Context) error {
+	if u.transient != "" {
+		_ = os.Remove(u.transient)
+	}
+	file, err := os.CreateTemp("dagstore", "transient")
```
aarshkshah1992, line 105: nit: should create the temp after the below `Mount.Stat` piece of code is successful.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+			s.state = ShardStateServing
+			s.refs++
+			go d.acquireAsync(tsk.Resp, s, s.mount)
+
+		case OpShardRelease:
+			if (s.state != ShardStateServing && s.state != ShardStateErrored) || s.refs <= 0 {
+				log.Warnf("ignored illegal request to release shard")
+				break
+			}
+			s.refs--
```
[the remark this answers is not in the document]
raulk: We use refcounts to know how many consumers are accessing shards. Closing an accessor should decrement the refcount (see #27) by sending an `OpShardRelease` task to the event loop.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+					_ = d.queueTask(&Task{Op: OpShardFail, Shard: tsk.Shard, Error: err}, d.completionCh)
+					return
+				}
+				_ = reader.Close()
+				_ = d.queueTask(&Task{Op: OpShardFetchDone, Shard: tsk.Shard}, d.completionCh)
+			}(d.ctx, s.mount)
+
+		case OpShardFetchDone:
+			s.state = ShardStateFetched
+			if !s.indexed {
```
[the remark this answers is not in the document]
raulk: I conceived the `Fetching` state as a general purpose state for when the shard's contents are being fetched from the mount. That means that when the shard has been registered completely, but the local transient goes missing, and a request is made, the shard would move to this state. Not sure if we'll end up using it this way after all. We can revisit later.

dagstore.go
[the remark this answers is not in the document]
raulk: I modelled them as separate states for debugging and restartability purposes (the system could restart after the content has been fetched, but before it has been indexed, and we shouldn't be needing to refetch). That said, things _could_ be simplified given that the local mirroring has been implemented in the mount.Upgrader, and not in the DAG store.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+					Key:   s.key,
+					Error: fmt.Errorf("failed to acquire shard: %w", tsk.Error),
+				}
+				for _, acqCh := range s.wAcquire {
+					go func(ch chan ShardResult) { ch <- res }(acqCh)
+				}
+				s.wAcquire = s.wAcquire[:0] // empty acquirers.
+			}
+
+			// TODO trigger retries?
```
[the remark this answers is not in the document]
raulk: Failure management is a TODO. We definitely don't want to delete the shard. We'd want to notify the application so they can retry.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+
+func (s *Shard) failShard(err error) {
+	s.state = ShardStateErrored
+	s.err = err
+}
+
+type OpType int
+
+const (
+	OpShardRegister OpType = iota
```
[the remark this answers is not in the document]
raulk: I'd rather keep them exported because it enables for tracing/notification use cases.

mount/upgrader.go
```diff
@@ -0,0 +1,160 @@
...
+			ret.transient = initial
+			return ret, nil
+		}
+	}
+
+	return ret, nil
+}
+
+func (u *Upgrader) Fetch(ctx context.Context) (Reader, error) {
+	if u.passthrough {
```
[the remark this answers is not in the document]
raulk: This is a bit of a dilemma. There could be remote mounts that allow for random access, although admittedly that's probably an inefficient approach for most access patterns, but not single-blok reads, for example. I think we'll want options in the `mount.Upgrader` to control under which circumstances we'll effectively upgrade.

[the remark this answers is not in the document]
raulk: https://github.com/filecoin-project/dagstore/issues/30

mount/upgrader.go
raulk: This needs to change anyway; we shouldn't be using temp files: #26.

mount/upgrader.go
raulk: What's the rationale?

mount/upgrader.go
raulk: Applied in fb333d8.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+	ShardStateErrored
+
+	// ShardStateUnknown indicates that it's not possible to determine the state
+	// of the shard, because an internal error occurred.
+	ShardStateUnknown
+)
+
+// Shard encapsulates the state of a shard inside the DAG store.
+type Shard struct {
+	sync.RWMutex
```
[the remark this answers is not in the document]
raulk: Yep, that's likely possible now.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+
+	state   ShardState
+	err     error // populated if shard state is errored.
+	indexed bool
+
+	refs uint32 // count of DAG accessors currently open
+
+}
+
+func (s *Shard) failShard(err error) {
```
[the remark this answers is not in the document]
raulk: Removed.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	dagst := &DAGStore{
+		mounts:       mounts,
+		config:       cfg,
+		indices:      indices,
+		shards:       make(map[shard.Key]*Shard),
+		externalCh:   make(chan *Task, 128), // len=128, concurrent external tasks that can be queued up before putting backpressure.
+		internalCh:   make(chan *Task, 1),   // len=1, because eventloop will only ever stage another internal event.
+		completionCh: make(chan *Task, 64),  // len=64, hitting this limit will just make async tasks wait.
```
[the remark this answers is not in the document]
raulk: We don't _need_ it right now, but it separates concerns nicely, and makes the code more future proof in case we do need to process tasks differently going forward.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+				res := ShardResult{Key: s.key}
+				go func(ch chan ShardResult) { ch <- res }(s.wRegister)
+				s.wRegister = nil
+			}
+
+			s.state = ShardStateAvailable
+
+			// trigger queued acquisition waiters.
+			for _, acqCh := range s.wAcquire {
+				go d.acquireAsync(acqCh, s, s.mount)
```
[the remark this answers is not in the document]
raulk: Well spotted!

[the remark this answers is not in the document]
raulk: a91f307

dagstore.go
raulk: Gotta analyse the tradeoffs carefully.

dagstore.go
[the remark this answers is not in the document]
raulk: Done.

dagstore.go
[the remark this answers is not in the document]
raulk: #33.

dagstore.go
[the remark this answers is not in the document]
raulk: This has been simplified now.

dagstore.go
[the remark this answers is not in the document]
raulk: This has been simplified now.

dagstore.go
```diff
@@ -0,0 +1,609 @@
...
+				_ = d.queueTask(&Task{Op: OpShardIndex, Shard: tsk.Shard}, d.internalCh)
+				break
+			}
+			// shard is indexed, we're ready to serve requests.
+			_ = d.queueTask(&Task{Op: OpShardMakeAvailable, Shard: tsk.Shard}, d.internalCh)
+
+		case OpShardIndex:
+			s.state = ShardStateIndexing
+			go func(ctx context.Context, mnt mount.Mount) {
+				reader, err := mnt.Fetch(ctx)
```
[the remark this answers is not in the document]
raulk: I'm not sure we want to do this for async goroutines, as not all operations are cancellable (e.g. CARv2 stuff).

raulk:
We have enough feedback here. I'm going to merge this PR and we can iterate on it. There's plenty to do still.

Issue #47, opened 2021-07-07
remove `indexed` shard state flag


raulk:
Replace with a call to `StatFullIndex()` to check if the index exists. This check is necessary when restoring a non-initialized shard after a restart.

Issue #27, opened 2021-07-05
implement shard release (ShardAccessor#Close)


raulk:
This should fire an `OpShardRelease` task to the event loop.

Issue #26, opened 2021-07-05
implement transient file tracking and cleanup


raulk:
We use a `mount.Upgrader` to mirror remote shards into local files. Those local files are currently (a) regular temporary files, and they are (b) not managed. We need to do the following in the upgrader:

- pass in the transient root directory to upgraders, so that transients are created inside that directory (referred to as "scratch space")
- replace the usage of temporary files with regular files under that directory.
- add a method `DeleteTransient() error` in the upgrader that deletes the transient file under the lock.

In the DAG store management layer, we need to do the following:
- expose a method `GC() error` that deletes the transient on all shards whose refcount is 0 (unused, i.e. not serving).

raulk:
The `DAGStore#GC() error` method was not been implemented in #49. Reopening.

PR #49, opened 2021-07-08
Fix Transient file cleanup, support CARv1 and CARv2 and remove the `indexed` field from Shard State

aarshkshah1992:
Closes https://github.com/filecoin-project/dagstore/issues/27.
~Closes https://github.com/filecoin-project/dagstore/issues/26.~
Closes https://github.com/filecoin-project/dagstore/issues/47.
Closes https://github.com/filecoin-project/dagstore/issues/48.

_**In addition to the above:**_
- Transient file cleanup MUST be based on transient file ref-counting, not on Shard access refcounting. We can run into all sorts of hairy bugs otherwise such as the async go-routines failing to decrement the refcount if they fail, not incrementing the refcount during shard registartion even though we do access the transient for Indexing etc etc.
- Fixes a bug where we were always expecting Mounts to ONLY return CARv2 files. We now support both CARv1 and CARv2.
- Fixes shard access refcounting -> we weren't releasing refcounts for when the shard acquire go-routines failed.
- Fixes nil value access exception during spinning up the initialisation go-routine to re-register an initialising shard upon resumption.
- Removes the `ShardStateServing ` state. This information can be discerned from the shard refcount and it's one less thing for us to keep track of/update correctly.
- The FS mount should support resumption by serializing the entire file path and NOT just the file name. Need this for resumption tests to work.
- Fixed a **hairy** bug that was blocking the entire event loop on a `mount.Fetch` operation. Basically shard persistence that happens in the event loop needs to take the Upgrader lock to get the transient file path but the underlying mount fetch also needs to take that look to ensure we fetch only once... Boom ! Have moved to a deterministic naming scheme for upgrader files and we lookup files for a given shard key on resumption rather than persisting the transient file path to the shard state.
- Introduces solid tests for  shard access refcount, dagstore resumption, blocking mounts and upgrader file refcounts.

### TODO
- [x] Solid tests for the Upgrader.

- We need to talk about failure management and recovery. I saw a case where we can block acquirers forever if we arent careful. Have commented on this PR at the relevant code line. I'm sure there are more bugs hiding in there.


accessor.go
```diff
@@ -26,21 +25,18 @@ type ReadBlockstore interface {
 // ShardAccessor provides various means to access the data contained
 // in a shard.
 type ShardAccessor struct {
-	key  shard.Key
-	data *car.Reader
-	idx  index.Index
+	key   shard.Key
```
raulk, line 28: Shard key is no longer necessary if we keep a reference to the shard.

accessor.go
```diff
@@ -49,12 +45,16 @@ func (sa *ShardAccessor) Shard() shard.Key {
...
+	bs, err := blockstore.NewReadOnly(sa.data, sa.idx)
 	return bs, err
 }
 
 // Close terminates this shard accessor, releasing any resources associated
 // with it, and decrementing internal refcounts.
 func (sa *ShardAccessor) Close() error {
-	return sa.data.Close()
+	if err := sa.data.Close(); err != nil {
+		return fmt.Errorf("failed to close shard: %w", err)
```
raulk, line 56: This error should be logged, but it should not prevent the release from happening.

dagstore.go
```diff
@@ -159,19 +159,14 @@ func NewDAGStore(cfg Config) (*DAGStore, error) {
...
-
 		if s.state == ShardStateInitializing {
 			// if we already have the index for the shard, there's nothing to do.
 			if istat, err := dagst.indices.StatFullIndex(s.key); err == nil && istat.Exists {
 				s.state = ShardStateAvailable
 			} else {
 				// restart the registration.
 				s.state = ShardStateNew
-				_ = dagst.queueTask(&task{op: OpShardRegister, shard: s}, dagst.externalCh)
+				_ = dagst.queueTask(&task{op: OpShardRegister, shard: s, waiter: &waiter{ctx: ctx}}, dagst.externalCh)
```
aarshkshah1992, line 169: ShardStateServing

dagstore_control.go
```diff
@@ -74,35 +76,39 @@ func (d *DAGStore) control() {
...
+				// to the event loop if it fails to acquire the shard.
 				s.refs++
 				go d.acquireAsync(tsk.ctx, w, s, s.mount)
 			}
 			s.wAcquire = s.wAcquire[:0]
 
 		case OpShardAcquire:
 			w := &waiter{ctx: tsk.ctx, outCh: tsk.outCh}
-			if s.state != ShardStateAvailable && s.state != ShardStateServing {
+			// TODO What if Shard is in the ShardStateErrored state  here ? For now, that will block the acquirer as
```
aarshkshah1992, line 88: @raulk The failure management/recovery mechanism will have to handle these cases.

raulk: The acquire should fail immediately, returning an error in the response channel that wraps the shard's error.

raulk (commented):
> Transient file cleanup MUST be based on transient file ref-counting, not on Shard access refcounting. We can run into all sorts of hairy bugs otherwise such as the async go-routines failing to decrement the refcount if they fail, not incrementing the refcount during shard registartion even though we do access the transient for Indexing etc etc.

Why do double refcounting, instead of only deleting transients from shards that are in `ShardStatusAvailable` with refcount=0 (i.e. not being used), or `ShardStatusErrored`?

> Fixes a bug where we were always expecting Mounts to ONLY return CARv2 files. We now support both CARv1 and CARv2.

Not a bug, just an unimplemented feature.

raulk (changes requested):
> Removes the ShardStateServing state. This information can be discerned from the shard refcount and it's one less thing for us to keep track of/update correctly.

Please roll this back. It may seem redudant, but:

1. It's a lot simpler to check one thing than to check two things to get the full picture.
2. "Serving" _is_ a FSM state; different decisions are taken based on whether the shard is being consumed or not (e.g. clearing transients). The more fine-grained and detailed an FSM can be, the better, as it allows for future extensibility.
3. Refcounts are not exposed publicly through `AllShardsInfo`, and they shouldn't, as they're an internal detail. But it's perfectly reasonable for callers to know if a shard is being actively used or not. By removing this state, you also removed the ability to expose that information through the API.

mount/fs.go
```diff
@@ -55,6 +57,9 @@ func (f *FSMount) Serialize() *url.URL {
 	if st, err := fs.Stat(f.FS, f.Path); err != nil {
 		u.Host = "irrecoverable"
 	} else {
+		q := u.Query()
+		q.Set(path, f.Path)
+		u.RawQuery = q.Encode()
 		u.Host = st.Name()
```
raulk, line 63: Basically what you discovered here is that `st.Name() != f.Path`. I was going on that assumption, but if that's not the case, let's just set the path in the URL host, instead of `st.Name()`

dagstore_async.go
```diff
@@ -43,40 +58,24 @@ func (d *DAGStore) acquireAsync(ctx context.Context, w *waiter, s *Shard, mnt mo
...
-		_ = d.queueTask(&task{op: OpShardFail, shard: s, err: err}, d.completionCh)
+		_ = d.failShard(s, fmt.Errorf("failed to acquire reader of mount: %w", err), d.completionCh)
 		return
 	}
-
 	defer reader.Close()
-
-	carreader, err := car.NewReader(reader)
+	// works for both CARv1 and CARv2.
+	idx, err := car.ReadOrGenerateIndex(reader)
```
raulk, line 66: This is a new API, and it's good for convenience. But we will not what to use it this way, because it's opaque and that harms observability and extensibility.

1. It could take 1 second if an inline index exists, or it could take minutes if this is forced to index a 64GiB CAR. And there's no way to know. Conversely, if you check the version and the header yourself, you can gather this information and expose it (through extra states like `ShardStateGeneratingIndex`, for example).
2. For better scalability, in the future we will want to control the scheduling of index generation. Imagine you plug in an external HDD with 1TiB worth of 4GiB CARs v1, each of which is a shard. With this API, you would end up indexing all of them at once, which will lead to a poor outcome due to the IO footprint of performing random access on 256 files at the same time.

dagstore_control.go
```diff
@@ -14,6 +14,8 @@ const (
 	OpShardAcquire
 	OpShardFail
 	OpShardRelease
+
+	opFlush // only for tests
```
raulk, line 18: This pattern is racy and leaky. And it's not really a flush because it's not technically _flushing_ anything. It's really inserting a "bookmark/breakpoint" in the event queue to get notified when it's hit. The problem is that this solution assuming that all events of interest have been executed _before_ the mark is inserted. That's why it's racy (they may not have), and leaky (as a developer, you need to know enough of the internals to know that everything interesting has executed before you insert the mark).

What you're trying to solve here in a low-key fashion is the lack of observability, which is useful both for production usage and testing.

Let's instead build a "notifier" mechanism, which you subscribe to from tests, but that can also be used by users to watch shard state transitions.

dagstore_control.go
```diff
@@ -55,7 +57,12 @@ func (d *DAGStore) control() {
 			}
 
 			s.state = ShardStateInitializing
-
+			// if we already have the index for this shard, there's nothing to do here.
+			if istat, err := d.indices.StatFullIndex(s.key); err == nil && istat.Exists {
+				_ = d.queueTask(&task{op: OpShardMakeAvailable, shard: s}, d.internalCh)
+				break
+			}
+			// otherwise, generate and persist an Index for the CAR payload of the given Shard.
```
raulk, line 65: 👍

dagstore_control.go
```diff
@@ -69,35 +76,39 @@ func (d *DAGStore) control() {
...
+			// optimistically increment the refcount to acquire the shard. The go-routine will send an `OpShardRelease` message
+			// to the event loop if it fails to acquire the shard.
 			s.refs++
-
 			go d.acquireAsync(tsk.ctx, w, s, s.mount)
 
 		case OpShardRelease:
-			if (s.state != ShardStateServing && s.state != ShardStateErrored) || s.refs <= 0 {
-				log.Warnf("ignored illegal request to release shard")
+			if (s.state != ShardStateAvailable && s.state != ShardStateErrored) || s.refs <= 0 {
```
raulk, line 106: Let's roll back the removal of the `Serving` state.

dagstore_test.go
```diff
@@ -80,17 +110,21 @@ func TestRegisterCarV1(t *testing.T) {
 	require.NoError(t, err)
 
 	res := <-ch
-	require.Error(t, res.Error)
-	require.Contains(t, res.Error.Error(), "invalid car version")
```
raulk, line 84: We need a test registering a shard with a corrupted CAR, as we've lost coverage for errors at registration.

dagstore_test.go
```diff
@@ -211,37 +236,42 @@ func TestConcurrentAcquires(t *testing.T) {
...
-	t.Run("256", func(t *testing.T) { acquireShard(t, dagst, k, 256) })
+	t.Run("1", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 1) })
+	t.Run("2", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 2) })
+	t.Run("4", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 4) })
+	t.Run("8", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 8) })
+	t.Run("16", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 16) })
+	t.Run("32", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 32) })
+	t.Run("64", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 64) })
+	t.Run("128", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 128) })
+	t.Run("256", func(t *testing.T) { acquireShardThenRelease(t, dagst, k, 256) })
```
raulk, line 246: Let's not overload test helpers. This specialises them too much to the extent that their utility becomes very narrow and diminished.

Can you please have acquire return `[]ShardAccessor`, and create a new test utility `release(...ShardAccessor)` that releases the accessors?

dagstore_test.go
```diff
@@ -211,37 +236,42 @@ func TestConcurrentAcquires(t *testing.T) {
...
 
 	info := dagst.AllShardsInfo()
 	require.Len(t, info, 1)
 	for _, ss := range info {
-		require.Equal(t, ShardStateServing, ss.ShardState)
+		require.Equal(t, ShardStateAvailable, ss.ShardState)
```
raulk, line 251: Roll back.

mount/fs.go
aarshkshah1992: Yeah, but then the URL parsing then fails.  Is there a way to put filepaths in URL hostnames ?

dagstore_async.go
aarshkshah1992: @raulk Sure, but can we make that change when we have to make that change ?

raulk (commented):
Unless necessary for technical reasons, let's move the test files back out to a root `testdata` dir, where they are more easily discoverable and reusable.

dagstore_async.go
aarshkshah1992: This is simple, makes for less code in the Sharded DAG Store that we are just getting off the ground and is completely untested in the wild and our Markets tests have just started showing green after integrating the Sharded DAG Store.

Please can we write code for the future use case when the future use case arrives ?

raulk: We already had code like this before, but I'll open an issue to track.

raulk: https://github.com/filecoin-project/dagstore/issues/50

dagstore_control.go
raulk: Taken care of in #51.

dagstore_control.go
raulk: `ShardStateServing` restored in #51.

dagstore_control.go
```diff
@@ -120,10 +131,11 @@ func (d *DAGStore) control() {
 				s.wAcquire = s.wAcquire[:0] // clear acquirers.
 			}
 
+			// TODO What about all those who have already acquired
```
raulk, line 134: This part of the code doesn't know which kind of error occurred. Could be that the index has disappeared, but active acquirers already have the index so we shouldn't interrupt them. If this is a physical error (e.g. shard data was physically deleted, or corrupted), we'll leave to the `ShardAccessor` (and the `ReadBlockstore`) to fail at some point. At that stage, the caller will call `ShardAccessor#Close` and eventually all active references will be released.

raulk: Documented in #51.

dagstore.go
```diff
@@ -262,12 +263,36 @@ func (d *DAGStore) AcquireShard(ctx context.Context, key shard.Key, out chan Sha
...
 	ShardState
 	Error error
+	refs  uint32
+}
+
+// GetShardInfo returns the current state of the registered shard with key k
+func (d *DAGStore) GetShardInfo(k shard.Key) ShardInfo {
+	d.lk.RLock()
+	defer d.lk.RUnlock()
+	s := d.shards[k]
```
raulk, line 291: This will panic if the shard does not exist!

raulk: Solved in #51.

dagstore_test.go
```diff
@@ -322,19 +439,65 @@ func acquireShard(t *testing.T, dagst *DAGStore, k shard.Key, n int) {
...
+	grp2, _ := errgroup.WithContext(context.Background())
+	for i := 0; i < n; i++ {
+		ac := accessors[i]
+		grp2.Go(func() error {
+			mu.RLock()
+			defer mu.RUnlock()
+			if err := ac.Close(); err != nil {
+				return err
+			}
+			return nil
```
raulk, line 480:
```suggestion
			return ac.Close()
```

dagstore_test.go
```diff
@@ -322,19 +439,65 @@ func acquireShard(t *testing.T, dagst *DAGStore, k shard.Key, n int) {
...
+	_, err := os.Stat(abs)
+	require.NoError(t, err)
+
+	// close all the accessors
+	grp2, _ := errgroup.WithContext(context.Background())
+	for i := 0; i < n; i++ {
+		ac := accessors[i]
+		grp2.Go(func() error {
+			mu.RLock()
+			defer mu.RUnlock()
```
raulk, line 476: Not needed at this point, as nobody is mutating the slice any longer.

raulk: Therefore the whole thing can be simplified to:

```go
grp2.Go(ac.Close)
```

mount/upgrader.go
```diff
@@ -43,11 +47,32 @@ func Upgrade(underlying Mount, rootdir, initial string) (*Upgrader, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	defer src.Close()
+
+	dst, err := os.Create(ret.transientFilePath())
+	if err != nil {
+		return nil, err
+	}
+	defer dst.Close()
```
raulk, line 70: I don't think we want a _fixed_ file per shard. This worked in your resumption test because you deleted the directory between shutdown and restart. Otherwise there's a 0-size file left behind which the DAG store thinks is a valid transient copy (because we don't do size or checksum checks at the moment), and it fails to initialise the shard. I think we need a safer approach, such that we don't expose the transient path (and it's a random file, so that it cannot be guessed) UNTIL we have finished fetching the mount the first time. Otherwise, the DAG store will save the transient path even if it hasn't finished downloading, and it will be corrupted on restart.

mount/upgrader.go
```diff
@@ -18,19 +21,20 @@ import (
 type Upgrader struct {
 	underlying  Mount
 	passthrough bool
+	shardKey    shard.Key
 
-	lk        sync.Mutex
-	transient string
-	rootdir   string
+	lk      sync.Mutex
+	refs    uint32
```
raulk, line 27: We really don't want to do refcounting at this level.

mount/upgrader.go
```diff
@@ -43,11 +47,32 @@ func Upgrade(underlying Mount, rootdir, initial string) (*Upgrader, error) {
...
+	if _, err := os.Stat(ret.transientFilePath()); err == nil {
+		return ret, nil
+	}
+	// clean up the existing transient if it's gone bad
+	_ = os.Remove(ret.transientFilePath())
+
+	if initial == "" {
+		return ret, nil
+	}
+	// we don't want to rename the file given by the client -> simply copy it and use the upgrader's file naming scheme here.
```
raulk, line 59: We don't need to rename anything, and we absolutely don't want to duplicate the file, as that is wasteful. We should just save the initial transient path as provided by the client, even if it's outside the transient root.

mount/upgrader.go
```diff
@@ -18,19 +21,20 @@ import (
...
 }
 
 var _ Mount = (*Upgrader)(nil)
 
 // Upgrade constructs a new Upgrader for the underlying Mount. If provided, it
 // will reuse the file in path `initial` as the initial transient copy. Whenever
 // a new transient copy has to be created, it will be created under `rootdir`.
-func Upgrade(underlying Mount, rootdir, initial string) (*Upgrader, error) {
-	ret := &Upgrader{underlying: underlying, rootdir: rootdir}
+func Upgrade(underlying Mount, rootdir string, shardKey shard.Key, initial string) (u *Upgrader, finalErr error) {
```
raulk, line 36: This shouldn't depend on package `shard`. Replacing the type of `key` for `string` is better.

mount/upgrader_test.go
```diff
@@ -1,9 +1,270 @@
...
+	key := shard.KeyFromString(fmt.Sprintf("%d", rand.Uint64()))
+	rootDir := t.TempDir()
+	u, err := Upgrade(mnt, rootDir, key, "")
+	require.NoError(t, err)
+	require.Zero(t, mnt.Count())
+
+	// now fetch in parallel
+	cnt := 20
+	var mu sync.Mutex
+	readers := make([]Reader, 0, cnt)
```
raulk, line 203: No need to protect with a mutex if you size the slice to len 20 and you update the corresponding positions in the goroutine.

mount/upgrader.go
```diff
@@ -150,22 +161,46 @@ func (u *Upgrader) refetch(ctx context.Context) error {
...
+func (u *Upgrader) transientFilePath() string {
+	return filepath.Join(u.rootdir, u.shardKey.String())
+}
+
+type transientReaderCloser struct {
+	Reader
+	u *Upgrader
+}
+
+func (m *transientReaderCloser) Close() error {
```
raulk, line 186: The reason we don't want to remove transients immediately when all acquirers have been closed is because this leads to inefficient space utilisation. Imagine you have 2TiB available for scratch space, and you receive 20 requests in a row (but not overlapping) for shard A, which weighs 64GiB. You will keep refetching and deleting those 64GiB over and over again, even if you had 97% unused space.

The transient management will probably work through an LRU/LFU cache with a size boundary that allows us to evict (delete) the least/eldest accessed shards on demand, as we need to reclaim space to serve new shards that we still don't have locally.

raulk:
Merging with my adjustments because I need to build on this.

mount/upgrader_test.go
```diff
@@ -1,9 +1,238 @@
...
+}
+
+func (c *countingMount) Fetch(ctx context.Context) (Reader, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.n++
+
+	return c.Mount.Fetch(ctx)
+}
```
raulk, line 151: This is incorrect, as the exclusive lock _also_ guards the mount fetch (thus we're not testing the deduplication of the mount!). This counter is better handled with an atomic int. Will change in a subsequent PR.

dagstore_async.go
aarshkshah1992: This makes a lot of sense to me.

dagstore_test.go
aarshkshah1992: @raulk We're accessing a loop variable in a go-routine here.

dagstore_control.go
aarshkshah1992: Makes a lot of sense.

dagstore_test.go
aarshkshah1992: Makes a lot of sense.

dagstore_control.go
aarshkshah1992: Makes sense.

mount/upgrader.go
aarshkshah1992: @raulk Yeah , I understand that. I implemented this as better GC is something we've planned for the future.

PR #51, opened 2021-07-10
review / adjustments to #49

raulk:
Review of https://github.com/filecoin-project/dagstore/pull/49.

Changes as individual commits.


raulk:
@aarshkshah1992 I'm going to merge because I need to continue building on this in separate PRs. Please do review and comment anyway.

dagstore_test.go
```diff
@@ -456,49 +493,43 @@ func acquireShardThenRelease(t *testing.T, dagst *DAGStore, k shard.Key, n int)
...
-	<-ch
-	info, err := dagst.GetShardInfo(key)
-	require.NoError(t, err)
-	return info
+// releaseAll releases all accessors for a given shard.
+func releaseAll(t *testing.T, dagst *DAGStore, k shard.Key, accs []*ShardAccessor) {
+	grp, _ := errgroup.WithContext(context.Background())
+	for _, acc := range accs {
+		// close all accessors.
+		grp.Go(acc.Close)
```
aarshkshah1992, line 473: Shouldn't we not pass a looping variable to a go-routine ?

aarshkshah1992:
@raulk I think we are leaking the transient file across and creating a new transient copy even if we already have one.

aarshkshah1992:
@raulk I think we are leaking existing transient files on restarts.

mount/upgrader.go
```diff
@@ -131,9 +131,13 @@ func (u *Upgrader) Close() error {
 }
 
 func (u *Upgrader) refetch(ctx context.Context) error {
-	_ = os.Remove(u.transientFilePath())
+	u.lk.Lock()
```
aarshkshah1992, line 134: @raulk 

1. If one refetch fails because of a transient issue , all contending go-routines that saw the same `sync.Once` will also fail to fetch the file. 

2. We don't clean up the temp file if fetching from the underlying Mount fails.

dagstore_test.go
raulk: Nope, because the usage of acc is not deferred, it's used immediately to obtain the reference to Close.

mount/upgrader.go
raulk: > If one refetch fails because of a transient issue , all contending go-routines that saw the same sync.Once will also fail to fetch the file.

Yes, this is what we want.

raulk: > We don't clean up the temp file if fetching from the underlying Mount fails.

Good point, absolutely need to do that cleanup.

raulk:
> I think we are leaking existing transient files on restarts because we aren't persisting the transient file path to the Shard state anymore.

Yep, that's correct. We need a way to reconcile files in the transients directory with transients actually referenced. Discrepancies can be deleted.

raulk:
Alternatively we could download into a temp file, then move the temp file to its final location insidse the transients dir.

raulk:
https://github.com/filecoin-project/dagstore/issues/58

PR #52, opened 2021-07-10
implement DAGStore#GC.

raulk:
Implements the `GC()` method outlined in https://github.com/filecoin-project/dagstore/issues/26.
Closes https://github.com/filecoin-project/dagstore/issues/26.

### DIFF
--- a/dagstore.go
+++ b/dagstore.go
@@ -10,6 +10,7 @@ import (
 	"github.com/filecoin-project/dagstore/index"
 	"github.com/filecoin-project/dagstore/mount"
 	"github.com/filecoin-project/dagstore/shard"
+	"github.com/hashicorp/go-multierror"
 	ds "github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
 	"github.com/ipfs/go-datastore/query"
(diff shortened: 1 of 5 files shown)


dagstore.go
```diff
@@ -331,6 +332,59 @@ func (d *DAGStore) AllShardsInfo() AllShardsInfo {
...
+// available but inactive.
+//
+// It is not strictly atomic for now, as it determines which shards to reclaim
+// first, sends operations to the event loop, and waits for them to execute.
+// In the meantime, there could be state transitions that change reclaimability
+// of shards (some shards deemed reclaimable are no longer so, and vice versa).
+//
+// However, the event loop checks for safety prior to deletion, so it will skip
+// over shards that are no longer safe to delete.
+func (d *DAGStore) GC(ctx context.Context) (map[shard.Key]error, error) {
```
aarshkshah1992, line 345: @raulk Please can we launch a go-routine to do this periodically (configurable period) in the DAG Store ? Otherwise, we'll have to place this burden on all clients.

dagstore.go
```diff
@@ -331,6 +332,59 @@ func (d *DAGStore) AllShardsInfo() AllShardsInfo {
...
+	var (
+		merr    *multierror.Error
+		reclaim []*Shard
+	)
+
+	d.lk.RLock()
+	for _, s := range d.shards {
+		s.lk.RLock()
+		if s.state == ShardStateAvailable || s.state == ShardStateErrored {
+			reclaim = append(reclaim, s)
```
aarshkshah1992, line 355: @raulk Since we have the Shard lock here, why not delete the transient here ? 

Deleting a transient needs the Upgrader lock and it dosen't mutate any Shard state, Why should the deletion of the transient only happen in the event loop if it dosen't lead us to updating the Shard state?

aarshkshah1992: That ay, we can also avoid the event loop having to take the Upgrader lock.

dagstore.go
raulk: I'm not convinced periodic GC is the best approach, although I filed it here: #56. For now, we probably want to keep it manual, but in the future we will want to monitor disk usage of the transients dir and perform GC when it reaches a watermark, not periodically.

dagstore.go
raulk: Because there could be acquire operations queued at this point that still haven't been reflected on the state.

raulk: Note: deleting a transient doesn't mutate the shard state _directly_ in the sense that it doesn't mutate the direct fields of the `Shard` object, but it does mutate the state in a more abstract manner.

aarshkshah1992: @raulk 

While I agree with the conceptual clarity point, please note that there is still a bug here:

- We spin a go-routine to acquire a shard. It fetches the transient .
- A GC runs -> enters the event loop -> deletes the transient as state is still available.
- The initial acquire go-routine completes and we return the broken accessor to the client.

The fix here is to optimistically mark Shard State as `Serving` before we spin up a go-routine to acquire a shard since we also optimistically increment the shard access refcount here.

aarshkshah1992: @raulk 

While I agree with the conceptual clarity point, please note that there is still a bug here:

- We spin a go-routine to acquire a shard. It fetches the transient .
- A GC runs -> enters the event loop -> deletes the transient as state is still available.
- The initial acquire go-routine completes and we return the broken accessor to the client.

The fix here is to optimistically mark Shard State as `Serving` before we spin up a go-routine to acquire a shard since we also optimistically increment the shard access refcount here.

aarshkshah1992: Filed https://github.com/filecoin-project/dagstore/issues/59.

raulk: There's a tiny window, close to impossible, but it's still there:

1. Async fetch finished and updated the transient path in the Upgrader.
2. Async fetch sends `OpShardMakeAvailable` to the event loop.
3. In parallel, GC marked this shard for reclaim.
4. **The `OpShardGC` arrives to the event loop _before_ the `OpShardMakeAvailable`.** Thus the shard hasn't been moved to `OpShardServing` yet, and its transient gets deleted.

The way to solve this is for `OpShardGC` to refuse to delete the transient if there are pending acquirers.

aarshkshah1992: If we are okay acquiring the refcount optimistically, why not mark the shard state as serving ? The shard state will be changed to Available later when the refcount drops to zero.

raulk: We actually do set the shard state to "Serving" optimistically, if it's in "Available" state. If this happens while initializing, and an acquirer comes midway, we park it but we don't update the state to "Serving", because the true main state is "Initializing".

PR #60, opened 2021-07-12
implement shard failure management and recovery

### LINKED ISSUE #33: implement shard failure management and recovery
Shards can fail due for various reasons:

1. mount is inaccessible
2. index failed to populate
3. internal incongruency
4. released with error

When a shard fails, the application must be notified and given the chance to recover.

## Implementation

1. Implement `RecoverShard(ctx context.Context, key shard.Key, out chan ShardResult) error`, which resets the state of the shard, removes its index, and re-registers it from scratch.
2. Accept an `ErrorCh chan ShardError` option in the constructor, which if not-nil will be notified of shard errors.

raulk:
Closes #33.

This implements failure monitoring and recovery. It introduces a new configuration option: `FailureCh` where the DAG Store will notify the application of shard failures. The application can then evaluate the error and decide if it wants to recover the shard.

Recovering the shard is done by calling:

```go
func (d *DAGStore) RecoverShard(ctx context.Context, key shard.Key, out chan ShardResult, _ RecoverOpts) error
```

The supplied channel is used to publish the recovery result back, and the supplied context governs the lifetime of the recovery.

I've turned the dispatch mechanism into a reusable primitive, which we now use to dispatch results AND failures (thus we don't block the event loop).

Unit testing is done by triggering indexing failures. We should probably add another test for fetch failures, but the code paths should be the same.


dagstore.go
```diff
@@ -50,22 +50,41 @@ type DAGStore struct {
...
 	// internalCh receives internal tasks to the event loop.
 	internalCh chan *task
 	// completionCh receives tasks queued up as a result of async completions.
 	completionCh chan *task
-	// dispatchCh is serviced by the dispatcher goroutine.
-	dispatchCh chan *dispatch
+	// dispatchResultsCh is a buffered channel for dispatching results back to
+	// the application. Serviced by a dispatcher goroutine.
+	// Note: This pattern decouples the event loop from the application, so a
+	// failure to consume immediately won't block the event loop.
```
aarshkshah1992, line 64: However, slow consumption by consumers can block the event loop if the dispatch channel gets full.

aarshkshah1992 (approved):
LGTM. No hairy bugs evident.

PR #66, opened 2021-07-14
mmap CARs when possible in ShardAccessors.

### LINKED ISSUE #28: mmap transients
mmap transient files using https://pkg.go.dev/golang.org/x/exp/mmap when passing them to the CARv2 library.

raulk:
Closes #28.

### DIFF
--- a/accessor.go
+++ b/accessor.go
@@ -2,6 +2,9 @@ package dagstore
 
 import (
 	"context"
+	"io"
+	"os"
+	"sync"
 
 	"github.com/filecoin-project/dagstore/mount"
 	"github.com/filecoin-project/dagstore/shard"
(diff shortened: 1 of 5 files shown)


accessor.go
```diff
@@ -27,6 +31,12 @@ type ShardAccessor struct {
 	data  mount.Reader
 	idx   index.Index
 	shard *Shard
+
+	// mmapr is an optional mmap.ReaderAt. It will be non-nil if the mount
+	// has been mmapped because the mount.Reader was an underlying *os.File,
+	// and an mmap-backed accessor was requested (e.g. Blockstore).
+	lk    sync.Mutex
```
raulk, line 38: Actually, this lock might not even be necessary because ShardAccessors aren't thread-safe because by definition because they operate on a reader.

PR #70, opened 2021-07-20
Ignore nil channels in the dispatcher

aarshkshah1992:
If the code directly writes to the channel that the `dispatcher` function is reading from, we could escape the nil channel check we've put in place in the `dispatchResult` function.

### DIFF
--- a/dagstore_test.go
+++ b/dagstore_test.go
@@ -51,6 +51,38 @@ func TestRegisterUsingExistingTransient(t *testing.T) {
 	require.NotNil(t, idx)
 }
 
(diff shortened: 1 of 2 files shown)


raulk (approved):

raulk (commented):
Should probably add a small test, though.

aarshkshah1992:
Have added a test and merging as CI is green.

Stebalien:
Please don't update to commits on random branches. We _delete_ random branches.

Stebalien:
Tagged https://github.com/ipld/go-car/releases/tag/v2.0.0-beta2 to revive the commit.

PR #75, opened 2021-07-21
Fix shard state string representation

aarshkshah1992:


dirkmc (approved):

raulk (commented):
Unfortunately, this PR replaces a cheap static array lookup with a dynamic map lookup, every time a string representation is needed.

What was broken is that the shard state int values were not linear, yet the array assumed they were. It's preferrable to solve this with a sparse array.

PR #78, opened 2021-07-21
add a failure handler utility to recover shards immediately.

raulk:
This introduces a utility that DAG Store users can rely on to recover shard failures automatically, up to a maximum number of attempts. It currently does not support backoff; see #79.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -11,7 +11,9 @@ import (
 // unique shard.
 //
 // Attempt tracking does not survive restarts. When the passed context fires,
-// the failure handler goroutine will yield
+// the failure handler will yield. It is recommended to call this
+// method from a dedicated goroutine, as it runs an infinite event
+// loop.
 func RecoverImmediately(ctx context.Context, dagst *DAGStore, failureCh chan ShardResult, maxAttempts uint64) {
 	var (
 		recResCh = make(chan ShardResult, 128)
--- a/handlers.go
+++ b/handlers.go
@@ -1,0 +1,59 @@
+package dagstore
+
+import (
+	"context"
+
+	"github.com/filecoin-project/dagstore/shard"
+)
+
+// RecoverImmediately takes a failureCh were DAGStore failures are sent, and
+// attempts to recover the shard immediately up until maxAttempts for each
+// unique shard.
+//
+// Attempt tracking does not survive restarts. When the passed context fires,
+// the failure handler goroutine will yield
+func RecoverImmediately(ctx context.Context, dagst *DAGStore, failureCh chan ShardResult, maxAttempts uint64) {
+	var (
+		recResCh = make(chan ShardResult, 128)
+		attempts = make(map[shard.Key]uint64)
+	)
+
+	for {
+		select {
+		case res := <-failureCh:
+			key := res.Key
+			att := attempts[key]
+			if att >= maxAttempts {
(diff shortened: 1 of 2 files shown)


handlers.go
```diff
@@ -0,0 +1,59 @@
...
+				log.Infow("failure handler: max attempts exceeded; skipping recovery", "key", key, "from_error", res.Error, "attempt", att)
+				continue
+			}
+
+			log.Infow("failure handler: recovering shard", "key", key, "from_error", res.Error, "attempt", att)
+
+			// queue the recovery for this key.
+			if err := dagst.RecoverShard(ctx, key, recResCh, RecoverOpts{}); err != nil {
+				log.Warnw("failure handler: failed to queue shard recovery", "key", key, "error", err)
+				continue
```
dirkmc, line 36: Here we bail out before incrementing the attempt count, so I think this will loop forever if RecoverShard always returns an error

handlers.go
```diff
@@ -0,0 +1,59 @@
...
+		case res := <-recResCh:
+			// this channel is just informational; a failure to recover will
+			// trigger another failure on the failureCh, which will be handled
+			// above for retry.
+			key := res.Key
+			if res.Error == nil {
+				log.Infow("failure handler: successfully recovered shard", "key", key)
+				delete(attempts, key)
+			} else {
+				log.Warnw("failure handler: failed to recover shard", "key", key, "attempt", attempts[key])
```
dirkmc, line 49: Should we be incrementing the attempts count here?

handlers.go
```diff
@@ -0,0 +1,59 @@
+package dagstore
+
+import (
+	"context"
+
+	"github.com/filecoin-project/dagstore/shard"
+)
+
+// RecoverImmediately takes a failureCh were DAGStore failures are sent, and
```
dirkmc, line 9:
```suggestion
// RecoverImmediately takes a failureCh where DAGStore failures are sent, and
```

handlers.go
```diff
@@ -0,0 +1,59 @@
...
+// attempts to recover the shard immediately up until maxAttempts for each
+// unique shard.
+//
+// Attempt tracking does not survive restarts. When the passed context fires,
+// the failure handler goroutine will yield
```
dirkmc, line 14: > the failure handler goroutine will yield

nit: this tripped me up the first time I read it - there is no go routine :)

raulk: Good point! I wrote this comment for a previous version ;-)

raulk:
```suggestion
// the failure handler will yield. It is recommended to call this
// method from a dedicated goroutine, as it runs an infinite event
// loop.
```

handlers.go
raulk: I perceive this behaviour as correct. Since our recovery request was **not accepted**, it cannot even be computed as "an attempt". This happens when a precondition is not met (e.g. shard does not exist), and we don't want to keep reattempting.

handlers.go
raulk: Not here, only when we actually trigger a recovery (this is the async result from a recovery, which if it fails, will _also_ trigger a failure, and we'll receive an event above, which will start another recovery attempt, if applicable)

handlers.go
aarshkshah1992: This makes sense to me though something is really messed up if we're attempting to recover a Shard that does NOT exist.

aarshkshah1992 (approved):

PR #81, opened 2021-07-22
Bug Fixes and logging

### LINKED ISSUE #80: option to recover failed shards on initialization
When restoring errored shards from the persistent store, we don't do anything special to try to recover them. We rely on the client to take action by querying all shard states and recovering failed shards. We should add an option on `dagstore.Config`:

```go
type Config struct {
    ...
    RecoverFailed bool
}
```

that, when true, automatically sends OpShardRecover tasks for all failed shards.

aarshkshah1992:
Closes #80 .

- If we persist a shard in errored state to the datastore and then crash and restart, the shard would have forever stayed in the errored state.

- Shard Recovery should respect lazy initialization.

- Document that clients have to always actively consume from the trace and failure channels.

- Logging for all the Upgrader magic. Logging for everything.

### TODO
- [x] GoDocs.
- [x] Solid Unit tests for the changes introduced here.

### DIFF
--- a/dagstore_control.go
+++ b/dagstore_control.go
@@ -51,6 +51,7 @@ func (d *DAGStore) control() {
 	for {
 		// consume the next task; if we're shutting down, this method will error.
 		if tsk, err = d.consumeNext(); err != nil {
+			log.Errorw("failed to consume next task in event loop, will return from event loop", "error", err)
 			break
 		}
 
@@ -265,6 +266,17 @@ func (d *DAGStore) control() {
 				log.Debugw("recovery: no index dropped for shard", "shard", s.key)
 			}
 
+			if s.lazy {
+				log.Debugw("shard recovered with lazy initialization, resetting shard state to ShardStateNew", "shard", s.key)
+				// waiter will be nil if this was a restart and not a call to RecoverShard().
+				if tsk.waiter != nil {
+					res := &ShardResult{Key: s.key}
+					d.dispatchResult(res, tsk.waiter)
+				}
+				s.state = ShardStateNew
+				break
+			}
+
 			// fetch again and reindex.
 			go d.initializeShard(tsk.ctx, s, s.mount)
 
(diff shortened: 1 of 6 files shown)


mount/upgrader.go
```diff
@@ -66,9 +72,13 @@ func (u *Upgrader) Fetch(ctx context.Context) (Reader, error) {
 	// after it's done, open the resulting transient.
 	u.lk.Lock()
 	if u.transient != "" {
+		log.Debugw("have an existing transient copy, checking if we can use it for the fetch", "shard", u.key, "transient", u.transient)
 		if _, err := os.Stat(u.transient); err == nil {
-			u.lk.Unlock()
+			log.Debugw("transient is still alive, not refetching", "shard", u.key, "transient", u.transient)
+			defer u.lk.Unlock()
```
aarshkshah1992, line 78: @raulk This was the unlock bug I was talking about. Hairy edge case that I think actually happened on the Sofia Miner.

dagstore_control.go
```diff
@@ -124,11 +126,13 @@ func (d *DAGStore) control() {
 				// optimistically increment the refcount to acquire the shard. The go-routine will send an `OpShardRelease` message
 				// to the event loop if it fails to acquire the shard.
 				s.refs++
+				// TODO Can we pass in  a looping pointer variable as is to a go-routine ?
```
aarshkshah1992, line 129: @raulk

raulk: This is fine. The pointer is passed by value. A different thing would be if we closed over the value, in which case it would be racy, e.g.:

```go
for _, s := range ... {
    go func() {
       // do something with s
    }()
}
```

Take a look at this: https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/.

raulk: When in doubt, you can cook up a little Go Playground thing: https://play.golang.org/p/TogtEsZAkiY

raulk (approved):

mount/upgrader.go
```diff
@@ -78,7 +88,12 @@ func (u *Upgrader) Fetch(ctx context.Context) (Reader, error) {
 	u.lk.Unlock()
 
 	var err error
-	once.Do(func() { err = u.refetch(ctx) })
+	once.Do(func() {
+		err = u.refetch(ctx)
+		if err != nil {
+			log.Errorw("failed to refetch", "shard", u.key, "error", err)
+		}
+	})
```
raulk, line 96: There's a bug here. This error won't be propagated to waiting fetchers.

PR #88, opened 2021-07-25
refactor GC and implement reconciliation.

### LINKED ISSUE #58: reconcile transients directory
If the system is restarted while a fetch is in progress, we will leave behind a partial, unreferenced file in the transients dir.

Implement a way to reconcile the actual contents with the transients directory with the referenced files.

Alternatively, the upgrader could download into a temp file, then move the file to its final location under the transients dir (probably the best option, and we can use mnt.Stat().Size to ensure we have enough space, or call GC otherwise).

raulk:
This PR closes #58.

It implements reconciliation of the transients directory, which consists of removing files that are no longer referenced by any mount, either as a complete or partial transient file.

We introduce the concept of a "partial file" in the Upgrader. The partial file is the file where a mount is being fetched into, while the fetch is in progress (cf. `.crdownload` in Chrome). This is not safe to expose as the `TransientPath`, because that would imply that it's fully functional. However, we don't want it saved in the Datastore.

I've also refactor the way that GC is triggered, so it's much neater now.

### DIFF
--- a/dagstore.go
+++ b/dagstore.go
@@ -233,6 +233,10 @@ func NewDAGStore(cfg Config) (*DAGStore, error) {
 		return nil, fmt.Errorf("failed to restore dagstore state: %w", err)
 	}
 
+	if err := dagst.clearOrphaned(); err != nil {
+		log.Warnf("failed to clear orphaned files on startup: %s", err)
+	}
+
 	// Reset in-progress states.
 	//
 	// Queue shards whose registration needs to be restarted. Release those
@@ -472,13 +476,8 @@ func (d *DAGStore) AllShardsInfo() AllShardsInfo {
 	return ret
 }
 
-// GC performs DAG store garbage collection.
-//
-// This comprises two steps:
-//  1. Reclaim the transient files of shards that are currently available but
-//  inactive, or errored.
-//  2. Reconciles the transients directory with the mounts, by removing orphaned
-//  files (unreferenced files).
+// GC performs DAG store garbage collection by reclaiming transient files of
+// shards that are currently available but inactive, or errored.
 //
 // GC runs with exclusivity from the event loop.
 func (d *DAGStore) GC(ctx context.Context) (*GCResult, error) {
--- a/dagstore.go
+++ b/dagstore.go
@@ -10,7 +10,6 @@ import (
 	"github.com/filecoin-project/dagstore/index"
 	"github.com/filecoin-project/dagstore/mount"
 	"github.com/filecoin-project/dagstore/shard"
-	"github.com/hashicorp/go-multierror"
 	ds "github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
 	"github.com/ipfs/go-datastore/query"
@@ -86,6 +85,8 @@ type DAGStore struct {
 	// back to the application. Serviced by a dispatcher goroutine.
 	// See note in dispatchResultsCh for background.
 	dispatchFailuresCh chan *dispatch
+	// gcCh is where requests for GC are sent.
+	gcCh chan chan *GCResult
 
 	// Channels not owned by us.
 	//
@@ -210,6 +211,7 @@ func NewDAGStore(cfg Config) (*DAGStore, error) {
 		internalCh:        make(chan *task, 1),       // len=1, because eventloop will only ever stage another internal event.
 		completionCh:      make(chan *task, 64),      // len=64, hitting this limit will just make async tasks wait.
 		dispatchResultsCh: make(chan *dispatch, 128), // len=128, same as externalCh.
+		gcCh:              make(chan chan *GCResult, 8),
 		traceCh:           cfg.TraceCh,
 		failureCh:         cfg.FailureCh,
 		throttleFetch:     noopThrottler{},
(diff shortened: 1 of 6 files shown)


shard_persist.go
```diff
@@ -22,8 +22,8 @@ type PersistedShard struct {
 }
 
 // MarshalJSON returns a serialized representation of the state. It must be
-// called from inside the event loop, as it accesses mutable state, or under a
-// shard read lock.
+// called with a shard lock (read, at lesat), such as from inside the event
```
aarshkshah1992, line 25:
```suggestion
// called with a shard lock (read, at least), such as from inside the event
```

mount/upgrader.go
```diff
@@ -96,11 +102,45 @@ func (u *Upgrader) Fetch(ctx context.Context) (Reader, error) {
 
 	once.Do(func() {
 		atomic.AddInt32(&u.fetches, 1)
-		err := u.refetch(ctx)
-		if err != nil {
-			log.Errorw("failed to refetch", "shard", u.key, "error", err)
+
+		// create a temporary file, partial file, and track it.
+		// onceErr can be set outside the lock because access is exclusive due to sync.Once
+		u.lk.Lock()
```
aarshkshah1992, line 108: What is this lock guarding ?

mount/upgrader.go
```diff
@@ -96,11 +102,45 @@ func (u *Upgrader) Fetch(ctx context.Context) (Reader, error) {
...
+		u.lk.Unlock()
+
+		// do the refetch; abort and remove/reset the partial if it fails.
+		u.onceErr = u.refetch(ctx, file)
+		if u.onceErr != nil {
+			log.Warnw("failed to refetch", "shard", u.key, "error", u.onceErr)
+			if err := os.Remove(u.partial); err != nil {
+				log.Warnw("failed to remove partial transient", "shard", u.key, "path", u.partial, "error", err)
+			}
+			u.partial = ""
```
aarshkshah1992, line 126: If the above lock is guarding the "partial" state, shouldn't we take a lock to mutate this ?

dagstore_gc.go
```diff
@@ -0,0 +1,101 @@
...
+		res.Shards[s.key] = err
+
+		// flush the shard state to the datastore.
+		if err := s.persist(d.config.Datastore); err != nil {
+			log.Warnw("failed to persist shard", "shard", s.key, "error", err)
+		}
+		s.lk.Unlock()
+	}
+
+	// B. Remove files that aren't referenced by the mounts.
```
aarshkshah1992, line 74: I don't think we need to do this reconciliation on every GC. Just doing it on startup should be fine. We'll have unreferenced files only if we crash before failing to persist the transient path or crash before removing a partial file on a fetch error or crash before we mark a partial as a transient. 

This is adding needless cognitive overhead as I just had to reason about how this could race with the Upgrader to convince myself that it wont.

dagstore.go
```diff
@@ -210,6 +211,7 @@ func NewDAGStore(cfg Config) (*DAGStore, error) {
 		internalCh:        make(chan *task, 1),       // len=1, because eventloop will only ever stage another internal event.
 		completionCh:      make(chan *task, 64),      // len=64, hitting this limit will just make async tasks wait.
 		dispatchResultsCh: make(chan *dispatch, 128), // len=128, same as externalCh.
+		gcCh:              make(chan chan *GCResult, 8),
```
aarshkshah1992, line 214: Are we just going ahead with a random number here ?

aarshkshah1992 (approved):

dagstore_control.go
```diff
@@ -36,32 +36,35 @@ func (o OpType) String() string {
...
 
-		log.Debugw("processing task", "op", tsk.op, "shard", tsk.shard.key, "error", tsk.err)
+		if gc != nil {
+			// this was a GC request.
+			d.gc(gc)
+			continue
+		}
 
 		s := tsk.shard
+		prevState := s.state
```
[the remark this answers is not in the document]
aarshkshah1992: Approved by mistake.

dagstore_control.go
```diff
@@ -341,27 +344,25 @@ func (d *DAGStore) control() {
...
 	}
 
 	select {
 	case tsk = <-d.externalCh:
-		return tsk, nil
+		return tsk, nil, nil
 	case tsk = <-d.completionCh:
-		return tsk, nil
+		return tsk, nil, nil
+	case gc := <-d.gcCh:
```
aarshkshah1992, line 363: This can race with the shard state book-keeping/reconciliation we do on startup because we submit ops to the external channel on startup.  Since the ordering inside a select is indeterminate, we could end up picking the gc channel here (if client calls GC after startup) over the requests queued in the external channel at startup.

I think it would be nice to maintain the property of finishing all ops the dagstore itself submits to the event loop on startup before we pick client requests to avoid shooting ourselves in the foot down the line.

aarshkshah1992:
@raulk Approved by mistake .

dagstore_gc.go
raulk: That makes sense. It would allow us to get rid of exposing the partial file path too, and the associated complexity. We had discussed this, but I completely forgot.

dagstore.go
raulk: Yep, if the channel is backlogged, the next GC request will make the GC() call wait.

dagstore_control.go
raulk: > I think it would be nice to maintain the property of finishing all ops the dagstore itself submits to the event loop on startup before we pick client requests to avoid shooting ourselves in the foot down the line.

We do that, by consuming from internalCh first.


> This can race with the shard state book-keeping/reconciliation we do on startup because we submit ops to the external channel on startup. Since the ordering inside a select is indeterminate, we could end up picking the gc channel here (if client calls GC after startup) over the requests queued in the external channel at startup.

I don't think that's wrong. In fact, I was considering the opposite: prioritising GC over external operations. The user may be executing GC because the system is under pressure, and if not run on time, other operations may fail.

dagstore_control.go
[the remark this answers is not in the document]
raulk: Yep.

dagstore_control.go
aarshkshah1992: @raulk 

Yes, but the problem is that the startup operations are being submitted to the external channel.

```
// release the queued registrations before we return.
	for _, s := range toRegister {
		_ = dagst.queueTask(&task{op: OpShardRegister, shard: s, waiter: &waiter{ctx: ctx}}, dagst.externalCh)
	}

	// queue shard recovery for shards in the errored state before we return.
	for _, s := range toRecover {
		_ = dagst.queueTask(&task{op: OpShardRecover, shard: s, waiter: &waiter{ctx: ctx}}, dagst.externalCh)
	}
```

Now that I think of it, these look like they should be submitted to the internal channel.

raulk: GC can't be called at that point because we haven't even returned the instance to the user.

raulk: The only thing that could happen is that operations remained queued when we return the DAGStore, and the user immediately calls GC. However, I'm not sure how that would be offensive.

aarshkshah1992: What I'm saying is that even after this request is queued,  if a user submits a GC request, it's possible for the GC request to be picked over the startup ops. This will break the property of "startup ops before all client calls".

aarshkshah1992: @raulk It's probably not offensive today. But, I'd like for us to maintain the invariant of all startup operations being executed before any clients ops are executed.

raulk: That's true for GC or non-GC operations. That property is not formalised. If we want to introduce that property (and I think it's a good idea), we should pass a chan ShardResult in the waiters, and wait until we've consumed all responses. However, we might not want to block the constructor if we do that (synchronous recovery can take long). The implied contract of a constructor is that it just constructs the object, it does not do heavy lifting. Instead, we may want to move this logic to a `Start` method.

aarshkshah1992: How about simply submitting all startup ops to the internal channel ? That too should give us this property, right ?

dagstore_gc.go
raulk: On second thought, I don’t think I want the reconciliation to run only on startup. It’s really valuable to have it run in normal GC, as it makes the system more resilient, e.g. imagine we’re stuck in a loop where fetching a 32GiB mount fails halfway, repeatedly. We’d keep accumulating garbage that would only get collected on startup. Also, in the future, we want to set a disk quota for the transients dir, and have GC kick off based on a watermark policy.

mount/upgrader.go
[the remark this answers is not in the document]
raulk: Simplified the lock handling.

mount/upgrader.go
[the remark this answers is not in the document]
raulk: We need locks because we modify the paths, which can be accessed from getters.

dagstore_gc.go
raulk: I've pushed a commit to simplify the solution. Since now we're using precomputed filenames for complete and partial files, and not generating random ones, the risk of us accumulating garbage due to thrash is very low/nil. So now we can rely on orphan cleanup on startup only.

aarshkshah1992 (approved):

PR #92, opened 2021-07-29
mount: add Ready in Stat; upgrader throttling when ready.

raulk:
- Atomically throttle fetch & copy in the Upgrader if underlying Mount is ready to serve the data.
- Do not throttle the fetch & copy if underlying Mount is not ready to prevent head of line blocking.
- Remove the top level fetch throttling outside the Upgrader to prevent head of line blocking.


dagstore.go
```diff
@@ -98,9 +98,8 @@ type DAGStore struct {
 
 	// Throttling.
 	//
-	throttleFetch throttle.Throttler
-	throttleCopy  throttle.Throttler
-	throttleIndex throttle.Throttler
+	throttleFetchCopy throttle.Throttler
```
raulk, line 101: I think we can just call this `throttleFetch` (as copy is part of the fetch operation).

dagstore_async.go
```diff
@@ -70,18 +59,7 @@ func (d *DAGStore) acquireAsync(ctx context.Context, w *waiter, s *Shard, mnt mo
...
-	var reader mount.Reader
-	err := d.throttleFetch.Do(ctx, func(ctx context.Context) error {
-		var err error
-		reader, err = mnt.Fetch(ctx)
-		if err == nil {
-			log.Debugw("initialize: successfully fetched from mount upgrader", "shard", s.key)
-		} else {
-			log.Warnw("initialize: failed to fetch from mount upgrader", "shard", s.key, "error", err)
-		}
-		return err
```
raulk, line 82: Why remove these logs?

dagstore_async.go
```diff
@@ -18,18 +18,7 @@ import (
...
 
-	var reader mount.Reader
-	err := d.throttleFetch.Do(ctx, func(ctx context.Context) error {
-		var err error
-		reader, err = mnt.Fetch(ctx)
-		if err == nil {
-			log.Debugw("acquire: successfully fetched from mount upgrader", "shard", s.key)
-		} else {
-			log.Warnw("acquire: failed to fetch from mount upgrader", "shard", s.key, "error", err)
-		}
```
raulk, line 29: Why remove these logs?

dagstore_test.go
```diff
@@ -583,75 +583,6 @@ func TestLazyInitialization(t *testing.T) {
 	require.EqualValues(t, 16, info.refs)
 }
 
-// TestThrottleFetch exercises and tests the fetch concurrency limitation.
-// Testing thottling on indexing is way harder...
-func TestThrottleFetch(t *testing.T) {
```
raulk, line 588: Thanks for the unit test at the upgrader level, but it would be great to keep the integration test too.

PR #93, opened 2021-08-02
vyzo's review of the dagstore

vyzo:
Haven't dived into the subdirectories yet.


vyzo:
So far one minor bug (see comments inline) and one recommendation for refactoring:
The task processing logic of the event loop should be extracted on its own method, which can then:
- lock/defer unlock to avoid having a really long lock scope with a bunch of pasta in-between
- return the next task and avoid the `internaclCh` indirection, which can now be removed altogether.

PR #95, opened 2021-08-05
honour context cancellation on acquire.

raulk:
If the acquirer cancelled the context, we would never deliver the accessor and never release the shard. This was visible when testing out the new migration logic on Lotus and interrupting the `lotus dagstore initialize-all` command.

### DIFF
--- a/dagstore.go
+++ b/dagstore.go
@@ -7,15 +7,16 @@ import (
 	"os"
 	"sync"
 
-	"github.com/filecoin-project/dagstore/index"
-	"github.com/filecoin-project/dagstore/mount"
-	"github.com/filecoin-project/dagstore/shard"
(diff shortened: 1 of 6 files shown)


dirkmc (approved):
Approving as this looks like it should fix the bug 👍 

It seems like there's quite a bit of repeated code in the error handling, I wonder if there's a way to simplify this somehow 🤔 . Ideally there would be a mechanism that would always decrement the ref-count no matter what happens. Could we add a defer function that checks for error, and if there's an error it releases the shard and dispatches the shard result?

raulk:
Agree. I think a minor refactor to reduce complexity/entropy in the acquire/release paths would make sense. Right now I would like to abstain from deduplicating because things still need to settle a little anyway, and I prefer to handle in a dedicated refactor spike.

PR #99, opened 2021-09-09
Inverted Index

aarshkshah1992:
This is the first draft of the top-level inverted Index in the DAGStore which will enable Markets to serve retrieval requests for any cid in a deal-dag and NOT just the payloadCid.

**Discussion points:**
   - There is a race between a deal expiring causing the corresponding entries being removed from the inverted Index and a client doing a lookup on the top-level Index. Do we live with this ?
 
**Blockers:** 
We need to block on a CARv2 Index iterator that the data-systems team is working on.


dagstore.go
```diff
@@ -318,6 +333,11 @@ func (d *DAGStore) Start(ctx context.Context) error {
 	return nil
 }
 
+// TODO: There is a race between calling this and deal expory which will cause deletion from the inverted index.
```
dirkmc, line 336: I think we decided we don't need to worry about this race now, right? We can probably just remove this comment

dagstore.go
```diff
@@ -7,6 +7,12 @@ import (
 	"os"
 	"sync"
 
+	"github.com/filecoin-project/go-indexer-core/store/memory"
```
dirkmc, line 10: nit: fix import ordering

invertedindex/core_index.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/filecoin-project/go-indexer-core"
+
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multihash"
+
+	"github.com/filecoin-project/dagstore/shard"
+)
+
+var ErrNotFound = errors.New("cid not found in Index")
```
dirkmc, line 15:
```suggestion
var ErrNotFound = errors.New("multihash not found in Index")
```

invertedindex/core_index.go
```diff
@@ -0,0 +1,123 @@
...
+		// go-indexer-core appends values to the existing values we already have for the key
+		// it also takes care of de-duplicating values.
+		_, err := d.is.Put(mh, valueForShardKey(s))
+		return err
+	})
+}
+
+func (d *IndexerCoreIndex) DeleteMultihashesForShard(sk shard.Key, mhIter MultihashIterator) error {
+	return mhIter.ForEach(func(mh multihash.Multihash) error {
+		// remove the given value i.e. shard key from the index for the given cid key.
```
dirkmc, line 40:
```suggestion
		// remove the given value i.e. shard key from the index for the given multihash
```

invertedindex/core_index.go
```diff
@@ -0,0 +1,123 @@
...
+		_, err := d.is.Remove(mh, valueForShardKey(sk))
+		return err
+	})
+}
+
+func (d *IndexerCoreIndex) GetShardsForCid(c cid.Cid) ([]shard.Key, error) {
```
dirkmc, line 46: I agree with Anton's suggestion that for this layer we should be consistent in only having multihashes in the interface, so this function would be
`GetShardsForMultihash(mh multihash.Multihash) ([]shard.Key, error)`

invertedindex/core_index.go
```diff
@@ -0,0 +1,123 @@
...
+func (d *IndexerCoreIndex) Size() (int64, error) {
+	return d.is.Size()
+}
+
+// TODO: Index till dosent's support this
+func (d *IndexerCoreIndex) NCids() (int64, error) {
+	return 0, nil
+}
+
+/*func (d *IndexerCoreIndex) Iterator() (Iterator, error) {
```
dirkmc, line 73: Let's just remove this iterator code, as discussed sync

dirkmc (approved):
LGTM 👍 

Just one outstanding nit about import ordering.

Also are we ok depending on these commit hashes (instead of releases)?
```
	github.com/ipld/go-car/v2 v2.0.3-0.20210910134712-2d2568297f26
	github.com/multiformats/go-multicodec v0.3.1-0.20210902112759-1539a079fd61
```

dagstore.go
```diff
@@ -63,12 +69,13 @@ var (
...
-	shards  map[shard.Key]*Shard
-	config  Config
-	indices index.FullIndexRepo
-	store   ds.Datastore
+	lk            sync.RWMutex
+	mounts        *mount.Registry
+	shards        map[shard.Key]*Shard
+	config        Config
+	indices       index.FullIndexRepo
+	invertedIndex invertedindex.Index
```
nonsense, line 77: I think we should export this field, so that we can call the Index API from Lotus.

nonsense: @aarshkshah1992 can we export the `invertedIndex` field?

nonsense (changes requested):
Blocking until we export the inverted index.

nonsense:
@dirkmc @aarshkshah1992 sorry, I don't have access to this repo. Please add the `invertedIndex` as exported field, in order to be able to interact with it from the CLI commands.

dagstore.go
raulk: Let's call this variable `topLevelIdx` (top level index).

dagstore.go
```diff
@@ -139,6 +148,8 @@ type Config struct {
 	// IndexRepo is the full index repo to use.
 	IndexRepo index.FullIndexRepo
 
+	InvertedIndex invertedindex.Index
```
raulk, line 151: Let's try to keep the level of godoc quality this code has.

dagstore_async.go
```diff
@@ -134,5 +136,32 @@ func (d *DAGStore) initializeShard(ctx context.Context, s *Shard, mnt mount.Moun
 		return
 	}
 
+	// add all cids in the shard to the inverted (cid -> []Shard Keys) index.
+	iterableIdx, ok := idx.(index.IterableIndex)
+	if !ok {
+		_ = d.failShard(s, d.completionCh, "shard index is not iterable")
+		return
+	}
```
raulk, line 144: Do we actually want to fail here? IIUC, the shard was initialized correctly and it's able to operate.

dagstore_async.go
```diff
@@ -134,5 +136,32 @@ func (d *DAGStore) initializeShard(ctx context.Context, s *Shard, mnt mount.Moun
...
+
+	mhIter := &mhIdx{iterableIdx: iterableIdx}
+	if err := d.invertedIndex.AddMultihashesForShard(mhIter, s.key); err != nil {
+		_ = d.failShard(s, d.completionCh, "failed to add shard multihashes to the inverted index: %w", err)
+		return
```
raulk, line 149: Same as above. Failing here makes the shard totally inaccessible, even if the caller has the shard ID.

invertedindex/index.go
```diff
@@ -0,0 +1,20 @@
...
+
+type Index interface {
+	AddMultihashesForShard(mhIter MultihashIterator, s shard.Key) error
+
+	DeleteMultihashesForShard(s shard.Key, mhIter MultihashIterator) error
+
+	GetShardsForMultihash(h multihash.Multihash) ([]shard.Key, error)
+
+	Size() (int64, error)
+}
```
raulk, line 20: 1. Let's call this type `index.Inverted`?
2. Please document the type and its methods -- let's keep the same level of godocs as we already have.

aarshkshah1992:
@nonsense Have pushed a commit to unexport.

invertedindex/core_index.go
```diff
@@ -0,0 +1,74 @@
+package invertedindex
```
raulk, line 1: Let's avoid package sprawl. Delete this package and move the types to the existing `index` package, with the proposed renames?

invertedindex/core_index.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/multiformats/go-multihash"
+
+	"github.com/filecoin-project/dagstore/shard"
+)
+
+var ErrNotFound = errors.New("multihash not found in Index")
+
+var _ Index = (*IndexerCoreIndex)(nil)
+
+type IndexerCoreIndex struct {
```
raulk, line 18: I think this type can be private. Try renaming `NewIndexerCore` to `index.NewInverted()`.

raulk (commented):
Mainly proposing a simplification in code structure and changes in error handling.

nonsense (approved):

dagstore_async.go
aarshkshah1992: Yes, but a database should be able to update all it's indices correctly before declaring an insert to be a success. We don't want to end up in a partial success case here where some indices have the latest and some do not.

raulk: That assumes that all accesses require querying the top-level index, which isn't the case.

PR #101, opened 2021-09-10
refactor inverse index to use multihashes instead of cids

dirkmc:


invertedindex/ds_index.go
```diff
@@ -137,7 +148,7 @@ func (d *DataStoreIndex) GetShardsForCid(c cid.Cid) ([]shard.Key, error) {
 	return shardKeys, nil
 }
 
-func (d *DataStoreIndex) NCids() (uint64, error) {
+func (d *DataStoreIndex) NMultihashes() (uint64, error) {
```
nonsense, line 151: Nit, but maybe calling this just `Length`? It is clear that this index contains `mh -> shards`, so `Length` sounds like it would return the number of items in the index.

```suggestion
func (d *DataStoreIndex) NMultihashes() (uint64, error) {
```

nonsense (approved):

invertedindex/ds_index.go
```diff
@@ -29,20 +30,20 @@ func NewDataStoreIndex(d ds.Batching) *DataStoreIndex {
...
+func (d *DataStoreIndex) AddMultihashesForShard(mhIter index.IterableIndex, s shard.Key) error {
 	batch, err := d.ds.Batch()
 	if err != nil {
 		return fmt.Errorf("failed to create ds batch: %w", err)
 	}
 
-	for _, c := range cidIter {
-		ck := ds.NewKey(c.String())
+	err = mhIter.ForEach(func(mh multihash.Multihash, _ uint64) error {
+		ck := ds.NewKey(mh.HexString())
```
aarshkshah1992, line 40: would be best to have a multiHashToDsKey function that all CRUD operations can use in this file.

aarshkshah1992 (commented):
Just one comment.

PR #105, opened 2021-11-10
Api to get iterable index from the DagStore

aarshkshah1992:
We should be able to get an iterable index out of the dagstore so the reference provider can serve the index to the network indexer.

### DIFF
--- a/dagstore.go
+++ b/dagstore.go
@@ -13,8 +13,6 @@ import (
 
 	"github.com/filecoin-project/go-indexer-core/store/memory"
 
-	"github.com/ipfs/go-cid"
-
 	"github.com/filecoin-project/dagstore/invertedindex"
 
 	ds "github.com/ipfs/go-datastore"
@@ -208,7 +206,7 @@ func NewDAGStore(cfg Config) (*DAGStore, error) {
 	}
 
 	if cfg.InvertedIndex == nil {
-		log.Warn("using in-memory inverted index")
+		log.Info("using in-memory inverted index")
 		cfg.InvertedIndex = invertedindex.NewIndexerCore(memory.New())
 	}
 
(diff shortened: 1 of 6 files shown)


dagstore.go
```diff
@@ -204,7 +208,7 @@ func NewDAGStore(cfg Config) (*DAGStore, error) {
 	}
 
 	if cfg.InvertedIndex == nil {
-		log.Info("using in-memory inverted index")
+		log.Warn("using in-memory inverted index")
```
raulk, line 211: Why warn?

dagstore.go
```diff
@@ -333,10 +337,28 @@ func (d *DAGStore) Start(ctx context.Context) error {
 	return nil
 }
 
+func (d *DAGStore) GetIterableIndex(key shard.Key) (carindex.IterableIndex, error) {
+	fi, err := d.indices.GetFullIndex(key)
+	if err != nil {
+		return nil, err
```
raulk, line 343: Wrap in "failed to get index"

dagstore.go
```diff
@@ -333,10 +337,28 @@ func (d *DAGStore) Start(ctx context.Context) error {
...
+}
+
 func (d *DAGStore) GetShardKeysForCid(c cid.Cid) ([]shard.Key, error) {
 	return d.invertedIndex.GetShardsForMultihash(c.Hash())
 }
 
+func (d *DAGStore) GetShardKeysForMultihash(h mh.Multihash) ([]shard.Key, error) {
+	return d.invertedIndex.GetShardsForMultihash(h)
+}
+
```
raulk, line 361: Method names are unspecific, as they do not imply what the relationship between the shard and CID/MH is. Please rename to "ShardsContainingCid" and "ShardsContainingMultihash"

(Not sure I like having two variants, either)

dirkmc (approved):
Modulo raul's suggestions LGTM 👍

dagstore.go
aarshkshah1992: Done.

dagstore.go
aarshkshah1992: Changed to having only the `ShardsContainingMultihash ` variant as we can always go from a cid -> multihash.

dagstore.go
aarshkshah1992: fixed.

PR #108, opened 2021-11-12
Change failure to log

aarshkshah1992:


dirkmc (approved):

dagstore_async.go
```diff
@@ -140,13 +140,13 @@ func (d *DAGStore) initializeShard(ctx context.Context, s *Shard, mnt mount.Moun
 	// add all cids in the shard to the inverted (cid -> []Shard Keys) index.
 	iterableIdx, ok := idx.(carindex.IterableIndex)
 	if !ok {
-		_ = d.failShard(s, d.completionCh, "shard index is not iterable")
+		log.Errorw("shard index is not iterable", "shard", s.key)
 		return
```
raulk, line 144: If we return here, the shard will never be made available, which defeats the purpose of this PR?

PR #111, opened 2021-12-11
Update carv2

arajasek:
This is on top of v0.4.2 because that's what lotus / markets is using (and the goal is to unblock Lotus upgrading all its deps)


whyrusleeping (approved):

arajasek:
Should also be sent into master eventually.

PR #116, opened 2022-01-21
Blockstore on all dagstore cids

aarshkshah1992:
For IPFS <-> Filecoin interop.

A blockstore that can serve retrieval for any cid across all shards that the dagstore has.

### TODO

- [ ] Add fuzzed test.
- [x] GetSize should use the underlying blockstore and not load the entire block in memory.
- [ ] When a blockstore is removed from the cache it is closed. Use reference counting to ensure that the blockstore is not closed while there is still an outstanding operation against the blockstore


readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// ---- we don't have a cached blockstore for a shard that can serve the block -> let's build one.
+
+	// select a valid shard that can serve the retrieval
+	sk, err := ro.shardSelectF(c, shards)
+	if err != nil {
+		return nil, fmt.Errorf("failed to select a shard: %w", err)
+	}
+
+	// load blockstore for the selected shard and tru to serve the cid from that blockstore.
```
dirkmc, line 122:
```suggestion
	// load blockstore for the selected shard and try to serve the cid from that blockstore.
```

readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+	case res = <-resch:
+		if res.Error != nil {
+			return nil, fmt.Errorf("failed to acquire shard %s: %w", sk, res.Error)
+		}
+	}
+
+	sa := res.Accessor
+	bs, err := sa.Blockstore()
+	if err != nil {
+		return nil, fmt.Errorf("failed top load read only blockstore for shard %s: %w", sk, err)
```
dirkmc, line 140:
```suggestion
		return nil, fmt.Errorf("failed to load read only blockstore for shard %s: %w", sk, err)
```

readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+	return &AllShardsReadBlockstore{
+		d:               d,
+		shardSelectF:    shardSelector,
+		blockstoreCache: bslru,
+		blockCache:      blkLru,
+	}, nil
+}
+
+func (ro *AllShardsReadBlockstore) Get(ctx context.Context, c cid.Cid) (b blocks.Block, finalErr error) {
+	logbs.Debugw("bitswap Get called", "cid", c)
```
dirkmc, line 70: I suggest removing the word "bitswap" from all of these log lines as it may not necessarily be bitswap that calls uses this read-only blockstore

readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+		return val.(blocks.Block), nil
+	}
+
+	// fetch all the shards containing the multihash
+	shards, err := ro.d.ShardsContainingMultihash(ctx, mhash)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch shards containing the block: %w", err)
+	}
+	if len(shards) == 0 {
+		return nil, errors.New("no shards contain the requested block")
```
dirkmc, line 90: I suggest creating a new publicly-accessible error for not-found errors. Probably bitswap will often query the blockstore for a block that it doesn't have, and Bitswap will want to return a different response to the client (DONT_HAVE) for not-found errors than for other kinds of errors. A bitswap implementation should be able to do something like
```
if xerrors.Is(err, dagstore.ErrBlockNotFound) { ... }
```

readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+	if len(shards) == 0 {
+		logbs.Debugw("bitswap Has: returning false no error", "cid", c)
+		return false, nil
+	}
+
+	_, err = ro.shardSelectF(c, shards)
+	if err != nil {
+		logbs.Debugw("bitswap Has error", "cid", c, "err", err)
+		return false, fmt.Errorf("failed to select a shard: %w", err)
+	}
```
dirkmc, line 175: Why do we need to run the shard select function here?

readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+		if !ok {
+			continue
+		}
+
+		rbs := val.(*accessorWithBlockstore).bs
+		blk, err := rbs.Get(ctx, c)
+		if err != nil {
+			ro.blockstoreCache.Remove(sk)
+			continue
+		}
```
dirkmc, line 106: Could you add a comment about why we should remove the blockstore with that shard if there's an error getting it?

readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, fmt.Errorf("failed top load read only blockstore for shard %s: %w", sk, err)
+	}
+
+	blk, err := bs.Get(ctx, c)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get block: %w", err)
+	}
+
+	// update the block cache and the blockstore cache
+	ro.blockstoreCache.Add(sk, &accessorWithBlockstore{sa, bs})
```
dirkmc, line 149: I'm wondering if we need the accessorWithBlockstore struct. Can we just store the accessor in the blockstore cache and call Blockstore() on the accessor each time we need the blockstore? Or does accessor.Blockstore() actually create a new blockstore?

readonly_bs.go
```diff
@@ -0,0 +1,207 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create lru cache for read only blockstores")
+	}
+
+	// instantiate the block cache
+	blkLru, err := lru.New(maxBlocks)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create lru cache for blocks: %w", err)
+	}
```
dirkmc, line 59: I don't think we need a block cache - if the user wants caching, they can just wrap the instance of AllShardsReadBlockstore in the existing blockstore cache:
https://github.com/ipfs/go-ipfs-blockstore/blob/master/caching.go

readonly_bs.go
aarshkshah1992: Because only blocks that has returns true for are retrievable.

Say you have configured this function to only allow deals which are available for free retrieval. Then has should return false if a given cid dosen't belong to any free deal so we're not asked to serve those blocks.

raulk (commented):
- How and when are shards released?
- Suggest naming as IndexBackedBlockstore.
- If possible, put in a different package (ex. indexbs?) as this is a utility and strictly not needed for the operation of the DagStore.
- How is the LRU cache configured?

raulk:
Is there a reason that this **_needs_** to be a member of the DagStore?

readonly_bs.go
aarshkshah1992: Yeah, there's a bunch of work that needs to be done to get a blockstore from an accessor.

Open/mmap the CARv1 payload and combine it with the CARv2 Index to get the read only blockstore. We should ideally do this only once.

readonly_bs.go
dirkmc: So then should we distinguish between an error and a selection of size zero?

readonly_bs.go
aarshkshah1992: This is done in the new `IndexBackedBlockstore ` type.

readonly_bs.go
aarshkshah1992: This is done in the new `IndexBackedBlockstore` type.

readonly_bs.go
aarshkshah1992: This is done in the new `IndexBackedBlockstore ` type.

readonly_bs.go
aarshkshah1992: This is done in the new `IndexBackedBlockstore ` type.

aarshkshah1992:
@raulk 

- Have renamed it as `IndexBackedBlockstore ` in a new `indexbs ` package.
- Yeah, I think it dosen't **need** to be a member of `dagstore`. Have moved it outside.
- Shards are released when they are evicted from the lru blockstore cache. See https://github.com/filecoin-project/dagstore/blob/8d9286a428a423f85648493c514dd47dddad5d5f/indexbs/indexbacked_bs.go#L44.

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,194 @@
...
+		blk, err := rbs.Get(ctx, c)
+		if err != nil {
+			// we know that the cid we want to lookup belongs to a shard with key `sk` and
+			// so if we fail to get the corresponding block from the blockstore for that shards, something has gone wrong
+			// and we should remove the blockstore for that shard from our cache.
+			ro.blockstoreCache.Remove(sk)
+			continue
+		}
+
+		// add the block to the block cache
```
dirkmc, line 97:
```suggestion
```

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,194 @@
...
+		logbs.Debugw("Get: returning from block store cache", "cid", c)
+		return blk, nil
+	}
+
+	// ---- we don't have a cached blockstore for a shard that can serve the block -> let's build one.
+
+	// select a valid shard that can serve the retrieval
+	sk, err := ro.shardSelectF(c, shards)
+	if err != nil {
+		return nil, ErrBlockNotFound
```
dirkmc, line 107: I think we may want to distinguish between a "not found" error, and other errors returned from shardSelectF

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	sa := res.Accessor
+	bs, err := sa.Blockstore()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load read only blockstore for shard %s: %w", sk, err)
+	}
+
+	blk, err := bs.Get(ctx, c)
+	if err != nil {
```
dirkmc, line 132: Let's check here if the block is not found, and if so, return `ErrBlockNotFound`

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,194 @@
...
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case res = <-resch:
+		if res.Error != nil {
+			return nil, fmt.Errorf("failed to acquire shard %s: %w", sk, res.Error)
+		}
+	}
+
+	sa := res.Accessor
+	bs, err := sa.Blockstore()
```
dirkmc, line 126: If it's an expensive operation to create the blockstore, should we use a lock here to make sure that it doesn't get created twice by parallel calls?

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if len(shards) == 0 {
+		logbs.Debugw("Has: returning false no error", "cid", c)
+		return false, nil
+	}
+
+	_, err = ro.shardSelectF(c, shards)
+	if err != nil {
+		logbs.Debugw("Has error", "cid", c, "err", err)
+		return false, ErrBlockNotFound
```
dirkmc, line 161: If the error is "not found" we should return `false, nil`

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,194 @@
...
+	logbs.Debugw("GetSize called", "cid", c)
+
+	blk, err := ro.Get(ctx, c)
+	if err != nil {
+		logbs.Debugw("GetSize error", "cid", c, "err", err)
+		return 0, fmt.Errorf("failed to get block: %w", err)
+	}
+
+	logbs.Debugw("GetSize success", "cid", c)
+	return len(blk.RawData()), nil
```
dirkmc, line 178: Is it possible to call `GetSize()` on the underlying blockstore instead of loading the block's raw data into memory?
Bitswap uses the `GetSize()` call [on the hot path](https://github.com/ipfs/go-bitswap/blob/b6f0cc7c83aaa27a39cc7e1b16ee34bba2d8b5b8/internal/decision/engine.go#L622) so it's important that it be as efficient as possible.

indexbs/indexbacked_bs.go
aarshkshah1992: The thing is, that should never happen because we've already checked that the shard contains the requested cid and so the bockstore for that shard should have the requested cid. Even if it is a `ErrBlockNotFound `, something has gone badly wrong here and so we shouldn't return a simple `ErrBlockNotFound ` here.

readonly_bs.go
aarshkshah1992: Done.

aarshkshah1992:
@raulk @dirkmc 

Hmm, there is a race here where two Get calls create two shard accesors and only one gets added to the lru cache and eventually relased whereas we lose the reference to the other one without ever releasing it (because multiple add calls to the lru cache for the same key only updates the priority of the item).

Working on fixing it.

indexbs/indexbacked_bs.go
aarshkshah1992: Done.

indexbs/indexbacked_bs.go
aarshkshah1992: Done.

indexbs/indexbacked_bs.go
aarshkshah1992: Done.

indexbs/indexbacked_bs.go
aarshkshah1992: Done.

aarshkshah1992:
@raulk Please can you take a look ?

indexbs/indexbacked_bs.go
aarshkshah1992: What do you mean by calling `GetSize()` on the  underlying blockstore?

indexbs/indexbacked_bs.go
dirkmc: Ok fair enough

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,232 @@
...
+
+	rbs := val.(*accessorWithBlockstore).bs
+	blk, err := rbs.Get(ctx, c)
+	if err != nil {
+		// we know that the cid we want to lookup belongs to a shard with key `sk` and
+		// so if we fail to get the corresponding block from the blockstore for that shards, something has gone wrong
+		// and we should remove the blockstore for that shard from our cache.
+		ro.blockstoreCache.Remove(sk)
+		return nil, err
+	}
```
dirkmc, line 210: IIUC, whenever we need to create a new blockstore, it will first call this function and the block won't exist, so `rbs.Get(ctx, c)` will return blockstore.ErrNotFound, and it will call `blockstoreCache.Remove(sk)`

aarshkshah1992: Not sure what you mean, this method first checks if we have a blockstore for the shard and returns if we don't. Note that we've already checked that the shard has the cid/multihash we are looking for before calling this method. 

I've changed the variable naming and added a comment to make this explicit.

indexbs/indexbacked_bs.go
```diff
@@ -0,0 +1,233 @@
...
+// IndexBackedBlockstore is a read only blockstore over all cids across all shards in the dagstore.
+type IndexBackedBlockstore struct {
+	d            *dagstore.DAGStore
+	shardSelectF ShardSelectorF
+
+	bsStripedLocks  [256]sync.Mutex
+	blockstoreCache *lru.Cache // caches the blockstore for a given shard for shard read affinity i.e. further reads will likely be from the same shard. Maps (shard key -> blockstore).
+}
+
+func NewIndexBackedBlockstore(d *dagstore.DAGStore, shardSelector ShardSelectorF, maxCacheSize int) (blockstore.Blockstore, error) {
```
aarshkshah1992, line 46: More offline feedback on what remains before we can productionize this:

1. A control mechanism similar to Graphsync (Given a request from cidX for peerY -> should I run it) ?
2. Rate limiting similar to Graphsync.
3. Be able to run Bitswap retrieval as a separate process.

Wondertan:
Hello. We would like to use DAGStore together with a top-level index. We also need a Blockstore over the top-level index to serve sampling requests to our light clients, so we would like to have this merged not to reinvent the wheel:
* What else needs to be done here to say it's done?
* Can we somehow facilitate the process?

Kindly ping @raulk, @dirkmc, and @aarshkshah1992. I know you are busy with FVM and Boost things, but this should not take too much time from you. Thanks!

willscott:
@hannahhoward probably has the most recent context here, and I think we are hoping to finish the work needed to get this finished.

dirkmc:
@Wondertan @MichaelMure we are hoping to get it merged this week

indexbs/indexbacked_bs.go
```diff
@@ -61,162 +64,220 @@ func NewIndexBackedBlockstore(d dagstore.Interface, shardSelector ShardSelectorF
...
+	block blocks.Block
+	size  int
+}
+
+func (ro *IndexBackedBlockstore) Get(ctx context.Context, c cid.Cid) (blocks.Block, error) {
+	res, err := ro.execOpWithLogs(ctx, c, BlockstoreOpGet)
+	if err != nil {
+		return nil, err
+	}
+	return res.block, err
```
aarshkshah1992, line 91: can we wrap the error ?

indexbs/indexbacked_bs.go
```diff
@@ -188,42 +179,13 @@ func (ro *IndexBackedBlockstore) execOp(ctx context.Context, c cid.Cid, op Block
...
 
-	// Call the operation on the blockstore
-	res, err = execOpOnBlockstore(ctx, c, sk, bs, op)
-	if err != nil {
-		return nil, err
-	}
-
-	// Update the blockstore cache
+	// Add the blockstore to the cache
 	ro.blockstoreCache.Add(sk, &accessorWithBlockstore{sa, bs})
```
aarshkshah1992, line 183: @dirkmc If multiple threads fail to get the blockstore above, they will ALL call Acquire Shard which in turn will increment the reference count for the shard access as many times. However, those reference counts will then not get released by calling Close on them(since that will only happen once on eviction). Thus, the other blockstores will leak.

aarshkshah1992: This is why the entire (Get + Add) i.e. _GetOrCreateIfNotExistAndGet_ dance needs to be atomic.

PR #123, opened 2022-06-07
Sugar syntax for acquiring a shard and logging the trace loop

aarshkshah1992:
Just some sugar syntax to synchronously acquiring a shard and for logging the trace loop in debug mode.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -62,3 +62,22 @@ func RecoverImmediately(ctx context.Context, dagst *DAGStore, failureCh chan Sha
 		}
 	}
 }
+
+// LogTraceLoop logs the output of the given trace channel until the given context expires.
+func LogTraceLoop(ctx context.Context, traceCh chan Trace, onDone func()) {
+	defer onDone()
+
+	for {
+		select {
+		// Log trace events from the DAG store
+		case tr := <-traceCh:
+			log.Debugw("trace",
+				"shard-key", tr.Key.String(),
+				"op-type", tr.Op.String(),
+				"after", tr.After.String())
+
+		case <-ctx.Done():
+			return
+		}
+	}
+}
--- a/helpers/shard.go
+++ b/helpers/shard.go
@@ -8,6 +8,9 @@ import (
 	"github.com/filecoin-project/dagstore/shard"
 )
 
+// AcquireShardSync blocks until the requested shard is acquired or the given context expires.
+// It returns the acquired shard if the acquire was successful. It is the caller's responsibility
+// to close the returned accessor when done.
 func AcquireShardSync(ctx context.Context, dagst *dagstore.DAGStore, sk shard.Key) (*dagstore.ShardAccessor, error) {
 	ch := make(chan dagstore.ShardResult, 1)
 
--- a/helpers/shard.go
+++ b/helpers/shard.go
@@ -1,0 +1,30 @@
+package helpers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/filecoin-project/dagstore"
+	"github.com/filecoin-project/dagstore/shard"
+)
+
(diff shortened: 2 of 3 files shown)


shard/key.go
```diff
@@ -30,6 +30,11 @@ func KeyFromCID(cid cid.Cid) Key {
 	return Key{str: cid.String()}
 }
 
+// KeyFromCIDMultihash returns a key based on the multihash of the given cid.
```
dirkmc, line 33: I'm curious, why do you need this method?

handlers.go
```diff
@@ -62,3 +62,22 @@ func RecoverImmediately(ctx context.Context, dagst *DAGStore, failureCh chan Sha
 		}
 	}
 }
+
+// LogTraceLoop logs the output of the given trace channel until the given context expires.
+func LogTraceLoop(ctx context.Context, traceCh chan Trace, onDone func()) {
+	defer onDone()
```
dirkmc, line 68: I'm not sure I understand why you need the onDone method here. Instead of passing this function as a parameter, can't the caller just do:
```
LogTraceLoop(ctx, ch)
onDone()
```

shard/key.go
aarshkshah1992: @dirkmc In RM, I wanna key all shards by their multihash i.e. if I have a CAR file for a given multihash, it dosen't matter what the codec is. This is just a helper method to create the shard key based on the multihash.

handlers.go
```diff
@@ -62,3 +62,22 @@ func RecoverImmediately(ctx context.Context, dagst *DAGStore, failureCh chan Sha
 		}
 	}
 }
+
+// LogTraceLoop logs the output of the given trace channel until the given context expires.
+func LogTraceLoop(ctx context.Context, traceCh chan Trace, onDone func()) {
```
raulk, line 67: This doesn't belong here. As a utility, this makes way too many assumptions: log level, message, etc. The leakage is obvious in the fact that you need to provide an `onDone` function as an argument.

helpers/shard.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/filecoin-project/dagstore/shard"
+)
+
+// AcquireShardSync blocks until the requested shard is acquired or the given context expires.
+// It returns the acquired shard if the acquire was successful. It is the caller's responsibility
+// to close the returned accessor when done.
+func AcquireShardSync(ctx context.Context, dagst *dagstore.DAGStore, sk shard.Key) (*dagstore.ShardAccessor, error) {
+	ch := make(chan dagstore.ShardResult, 1)
+
+	if err := dagst.AcquireShard(ctx, sk, ch, dagstore.AcquireOpts{}); err != nil {
```
raulk, line 17: To make this a true utility function with no assumptions built-in, you should pass `AcquireOpts` as an argument.

shard/key.go
```diff
@@ -30,6 +30,11 @@ func KeyFromCID(cid cid.Cid) Key {
...
+func KeyFromCIDMultihash(cid cid.Cid) Key {
```
raulk, line 34: Is it too hard to do this in the application? Everything else is a constructor. If anything, I'd make this a `KeyFromMultihash` and pass in the multihash directly.

raulk (changes requested):

handlers.go
aarshkshah1992: Removed this function completely.

handlers.go
aarshkshah1992: Removed this.

helpers/shard.go
aarshkshah1992: Done.

shard/key.go
aarshkshah1992: Removed this completely.

aarshkshah1992:
Review addressed and helper added in https://github.com/filecoin-project/dagstore/pull/128.

PR #124, opened 2022-06-15
[WIP] Automated LRU GC sketch PR

aarshkshah1992:
The problem with automated LRU GC is that it can also GC shards that are being served/initialised. This will fail the ongoing initialise/acquire/blockstore read operations for those shards. We simply look at the directory size and keep removing transients till we hit our target. We can launch with this for V0 and fix it down the line.

Clients who are not okay with interrupting existing shard acquires/initialisation can continue using the existing manual but safe GC mechanism for now.

It's hard to make the automated LRU GC thread safe with the async shard acquire and shard init ops as they write to the transient directory in their own go-routine outside the dagstore event loop. The way the existing manual GC mechanism gets away with this is by ignoring shards that are being initialised or served.

The other thing to note here is that we may not always have Mounts that can tell you the size of the CAR upfront because the CAR maybe the result of a remote (root, selector) traversal and the size will be known only upon fetching the subDAG as a result of a remote traversal.  One example of this is the IPFS Gateway which does NOT have a HEAD method to know the size of a (root, selector) CAR upfront. It only exposes a GET API to stream the result of a (root, selector) traversal but does NOT have a HEAD method to know the size upfront. This makes the reactive "evict before a Mount fetch" strategy hard to implement. The simple thing to do here is to always run the GC proactively in the dagstore event loop when we detect watermark breach.

### DIFF
--- a/dagstore_control.go
+++ b/dagstore_control.go
@@ -3,6 +3,7 @@ package dagstore
 import (
 	"context"
 	"fmt"
+	"time"
 )
 
 type OpType int
@@ -41,6 +42,15 @@ func (d *DAGStore) control() {
 	var wFailure = &waiter{ctx: d.ctx, outCh: d.failureCh}
 
 	for {
+		// do we need to do a GC ?
+		ts, err := d.transientDirSize()
+		if err != nil {
+			continue
+		}
+		if float64(ts) >= float64(d.config.MaxTransientSize)*d.config.TransientsGCWatermarkHigh {
+			d.lruGC(float64(d.config.MaxTransientSize) * d.config.TransientsGCWatermarkHigh)
+		}
+
 		// consume the next task or GC request; if we're shutting down, this method will error.
 		tsk, gc, err := d.consumeNext()
 		if err != nil {
@@ -52,11 +62,12 @@ func (d *DAGStore) control() {
 			return
 		}
 
-		if gc != nil {
+		// Disable manaual GC if client wants automated LRU GC.
+		/*if gc != nil {
 			// this was a GC request.
 			d.gc(gc)
 			continue
-		}
+		}*/
 
 		s := tsk.shard
 		log.Debugw("processing task", "op", tsk.op, "shard", tsk.shard.key, "error", tsk.err)
@@ -133,6 +144,7 @@ func (d *DAGStore) control() {
 			s.wAcquire = s.wAcquire[:0]
 
 		case OpShardAcquire:
+			s.lastAccessedAt = time.Now()
 			log.Debugw("got request to acquire shard", "shard", s.key, "current shard state", s.state)
 			w := &waiter{ctx: tsk.ctx, outCh: tsk.outCh}
 
--- a/dagstore_gc.go
+++ b/dagstore_gc.go
@@ -4,6 +4,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
 
 	"github.com/filecoin-project/dagstore/shard"
 )
@@ -48,7 +49,7 @@ func (d *DAGStore) gc(resCh chan *GCResult) {
 	}
 	d.lk.RUnlock()
 
-	// attempt to delete transients of reclaimed shards.
+	// attempt to delete transients of reclaimed shards and free up required amount space.
 	for _, s := range reclaim {
 		// only read lock: we're not modifying state, and the mount has its own lock.
 		s.lk.RLock()
@@ -71,6 +72,80 @@ func (d *DAGStore) gc(resCh chan *GCResult) {
 	case resCh <- res:
 	case <-d.ctx.Done():
 	}
+}
+
+// gc performs DAGStore GC. Refer to DAGStore#GC for more information.
+//
+// The event loops gives it exclusive execution rights, so while GC is running,
+// no other events are being processed.
+func (d *DAGStore) lruGC(target float64) {
+	// sort shards by LRU
+	d.lk.RLock()
+	var reclaim []*Shard
+	for _, s := range d.shards {
+		s.lk.RLock()
+		reclaim = append(reclaim, s)
+		s.lk.RUnlock()
+	}
+	d.lk.RUnlock()
+
+	sort.Slice(reclaim, func(i, j int) bool {
+		reclaim[i].lk.RLock()
+		defer reclaim[i].lk.RUnlock()
+
+		reclaim[j].lk.RLock()
+		defer reclaim[j].lk.RUnlock()
+
+		return reclaim[i].lastAccessedAt.Before(reclaim[j].lastAccessedAt)
+	})
+
+	// TODO What should we do about shards that are currently being initialised/acquired ?
+	// Removing the transient here will fail those operations
+	// The problem is those operations happen async outside the event loop.
+	// I think as a start, we should only enable automated LRU GC if client
+	// is okay with ongoing ops failing because of it i.e. "forced" mode.
+	// If client is NOT okay with the "forced" mode, client can fallback to the existing safe GC mechanism
+	// where we do NOT mess with shards that are being initialised or served.
+
+	// attempt to delete transients of reclaimed shards and free up required amount space.
+	for _, s := range reclaim {
+		// break when transient directory achieves required size
+		size, err := d.transientDirSize()
+		if err != nil {
+			continue
+		}
+		if float64(size) < target {
+			break
+		}
+
+		// only read lock: we're not modifying state, and the mount has its own lock.
+		// Is it okay to have a 'force' mode where it is okay to remove those transients and let the ops fail ?
+		s.lk.RLock()
+		err = s.mount.DeleteTransient()
+		if err != nil {
+			log.Warnw("failed to delete transient", "shard", s.key, "error", err)
+		}
+
+		// flush the shard state to the datastore.
+		if err := s.persist(d.ctx, d.config.Datastore); err != nil {
+			log.Warnw("failed to persist shard", "shard", s.key, "error", err)
+		}
+		s.lk.RUnlock()
+	}
+}
+
+func (d *DAGStore) transientDirSize() (int64, error) {
+	var size int64
+	err := filepath.Walk(d.config.TransientsDir, func(_ string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() {
+			size += info.Size()
+		}
+		return err
+	})
+	return size, err
 }
 
 // clearOrphaned removes files that are not referenced by any mount.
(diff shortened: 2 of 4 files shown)


dagstore_gc.go
```diff
@@ -73,6 +74,80 @@ func (d *DAGStore) gc(resCh chan *GCResult) {
...
+		reclaim[i].lk.RLock()
+		defer reclaim[i].lk.RUnlock()
+
+		reclaim[j].lk.RLock()
+		defer reclaim[j].lk.RUnlock()
+
+		return reclaim[i].lastAccessedAt.Before(reclaim[j].lastAccessedAt)
+	})
+
+	// TODO What should we do about shards that are currently being initialised/acquired ?
```
aarshkshah1992, line 102: @raulk

dagstore_control.go
```diff
@@ -41,6 +42,15 @@ func (d *DAGStore) control() {
 	var wFailure = &waiter{ctx: d.ctx, outCh: d.failureCh}
 
 	for {
+		// do we need to do a GC ?
+		ts, err := d.transientDirSize()
+		if err != nil {
+			continue
```
raulk, line 48: Log here.

dagstore_control.go
```diff
@@ -41,6 +42,15 @@ func (d *DAGStore) control() {
 	var wFailure = &waiter{ctx: d.ctx, outCh: d.failureCh}
 
 	for {
+		// do we need to do a GC ?
+		ts, err := d.transientDirSize()
```
raulk, line 46: `ts` usually stands for timestamp.

```suggestion
		dirSize, err := d.transientDirSize()
```

dagstore_control.go
```diff
@@ -41,6 +42,15 @@ func (d *DAGStore) control() {
...
+		if err != nil {
+			continue
+		}
+		if float64(ts) >= float64(d.config.MaxTransientSize)*d.config.TransientsGCWatermarkHigh {
+			d.lruGC(float64(d.config.MaxTransientSize) * d.config.TransientsGCWatermarkHigh)
+		}
```
raulk, line 52: Easier to read if you flip the condition to shortcircuit when it's not met.

raulk: Can use local if variables to avoid being pedantic.

dagstore_control.go
```diff
@@ -52,11 +62,12 @@ func (d *DAGStore) control() {
...
 		}
 
-		if gc != nil {
+		// Disable manaual GC if client wants automated LRU GC.
+		/*if gc != nil {
 			// this was a GC request.
 			d.gc(gc)
 			continue
-		}
+		}*/
```
raulk, line 70: Automated GC should not preclude manual GC requests from happening. Any garbage collector there (e.g. JVM, Go, etc.) will honour manual GC runs.

dagstore_control.go
```diff
@@ -133,6 +144,7 @@ func (d *DAGStore) control() {
 			s.wAcquire = s.wAcquire[:0]
 
 		case OpShardAcquire:
+			s.lastAccessedAt = time.Now()
```
raulk, line 147: Storing the timestamp here leaks the GC abstraction. Instead, make a GC abstraction that you `NotifyAccessed(Shard)`. That GC abstraction can be a noop (if no automated GC is enabled), it could be an LRU cache, an LFU cache, an ARC cache, etc. But this approach is not right as it leaks the LRU strategy into the shard's data model.

dagstore_gc.go
raulk, line 102: This leaks. This should be part of a LRUGarbageCollector.

raulk: By storing the access metadata in the garbage collector, you can likely remove much of the locking.

raulk: Potential interfaces.

If garbage collector manages the directory directly:

```go
type GarbageCollector interface {
    NotifyAccessed(*shard.Shard) // invoked every time an acquirer requests the shard
    NotifyReclaimable(*shard.Shard) // when shard goes back from Serving to Available; we could notify every time we release, but then the GarbageCollector would need to do usage tracking, which would be a duplication of responsibliities
    NotifyRemoved(*shard.Shard) // stop tracking this shard, e.g. when a shard is terminally destroyed and the dagstore has already removed its resources
    Reclaim() []*shard.Shard // runs GC and returns which shards were reclaimed
}
```

If garbage collector only advises which shards are reclaimable, but does not own the management of the transients dir (the dagstore does):

```go
type GarbageCollector interface {
    NotifyAccessed(shard.Key) // invoked every time an acquirer requests the shard
    NotifyReclaimable(shard.Key) // when shard goes back from Serving to Available; we could notify every time we release, but then the GarbageCollector would need to do usage tracking, which would be a duplication of responsibliities
    NotifyRemoved(shard.Key) // stop tracking this shard, e.g. when a shard is terminally destroyed and the dagstore has already removed its resources

    Reclaimable() []shard.Key // returns all shards ordered by reclaimability precedence
    NotifyReclaimed([]shard.Key) // informs which shards were ultimately reclaimed
}
```

dagstore_gc.go
```diff
@@ -73,6 +74,80 @@ func (d *DAGStore) gc(resCh chan *GCResult) {
...
+		s.lk.RLock()
+		err = s.mount.DeleteTransient()
+		if err != nil {
+			log.Warnw("failed to delete transient", "shard", s.key, "error", err)
+		}
+
+		// flush the shard state to the datastore.
+		if err := s.persist(d.ctx, d.config.Datastore); err != nil {
+			log.Warnw("failed to persist shard", "shard", s.key, "error", err)
+		}
```
raulk, line 132: Why do this for every shard? Can't we do it once after the loop ends?

raulk: (Need to ensure no corruption in case of an abrupt restart/interruption; however, that interruption can happen in the current code anyway, so we already need to be prepared for it)

raulk:
> The other thing to note here is that we may not always have Mounts that can tell you the size of the CAR upfront because the CAR maybe the result of a remote (root, selector) traversal and the size will be known only upon fetching the subDAG as a result of a remote traversal. One example of this is the IPFS Gateway which does NOT have a HEAD method to know the size of a (root, selector) CAR upfront. It only exposes a GET API to stream the result of a (root, selector) traversal but does NOT have a HEAD method to know the size upfront. This makes the reactive "evict before a Mount fetch" strategy hard to implement. The simple thing to do here is to always run the GC proactively in the dagstore event loop when we detect watermark breach.

**Some options:**

1. Enhance IPFS with with a two-stage export process.
    1. Stage 1 is the preparation of the export CAR. Runs the traversal, packs the CAR, and caches it on disk, returning metadata (e.g. size, node count, etc.) about the CAR/DAG and a unique export ID with which the client can start the download.
    2. Stage 2 is the download via the export ID.
2. Make GC streaming/interactive (assuming we don't get the above feature implemented in IPFS).
    1. As we are downloading the mount, somehow detect when we've run out of transient quota.
    2. Pause the download, run GC, resume the download.

Option 1 would be the most generally useful one. Option 2 implies extra complexity I'd like to avoid and it feels like a workaround.

aarshkshah1992:
@raulk I personally do not thing augmenting the IPFS Gateway with a HEAD protocol is the way to go here. Even Filecoin SPs do not expose an API to know the size of the CAR upfront as traversal also costs computing resources. 

If we force all mounts to report the size of the CAR upfront, the dagstore quickly becomes unusable for a lot of remote IPLD style (root, selector) Mounts because a lot of "servers" wont be okay paying the cost of a traversal/caching the traversal. This also introduces an additional hops in the retrieval flow.

I think the way to go here is giving users a config option where they can express that they are going to use Mounts that can not know the size upfront and therefore they are okay  for the transient directory to temporarily exceed the watermark.

raulk:
> Even Filecoin SPs do not expose an API to know the size of the CAR upfront as traversal also costs computing resources.
I'm not sure I follow -- Filecoin SPs don't even use transients?

> I think the way to go here is giving users a config option where they can express that they are going to use Mounts that can not know the size upfront and therefore they are okay for the transient directory to temporarily exceed the watermark.

This configuration value would be really hard to adjust. Temporarily exceeding the watermark is not an acceptable solution, because it defeats the purpose of quotas altogether.

aarshkshah1992:
@raulk I mean a Mount that can fetch from a  Filecoin SP.

aschmahmann:
> Enhance IPFS with with a two-stage export process

Agree with @aarshkshah1992 that this seems like a bad idea. Issues include:
1. Your time-to-first-byte now requires pre-processing the entire file which seems bad
2. If the gateway/export endpoint is required to keep the CAR file on disk/memory until it's served then you need to figure out:
    a. How long is the request valid for? How long do I keep the CAR file on disk/memory before timing out?
    b. Is there a maximum size CAR file I'm now willing to export as a function of needing to buffer everything? Previously we could just stream and GC the bytes.
    c.  We'd have to stream the bytes into the buffer which presumably will also have some sort of maximum size + garbage collection policy. It seems like the work you're trying to avoid doing in the dagstore around transient quotas and garbage collection would now be required of every IPFS gateway implementation. Basically option 1 implies that you're doing option 2 somewhere else in the stack which means none of the concerns with option 2 are resolved.

raulk:
@aarshkshah1992 @aschmahmann you're right. For some reason my brain assumed that go-ipfs (Kubo) was generating a CAR first and _then_ streaming it over. I _knew_ that wasn't the case, but I totally blanked out (I blame context switch!). Yeah, in that case a two-stage process definitely doesn't make sense.

I think something along the lines of the second solution is likely the way forward then. We'll need to do work as we stream the CAR. A potentially elegant solution could revolve around reservations and allocations:

1. When starting to download a transient, the Mount would ask the GarbageCollector for an allocation. We can either provide a size hint (power of two), or a default size (e.g. 128MiB).
2. As we are streaming the file, we would use an `io.Pipe` or similar to count the bytes we are receiving, and detecting when we're going above the awarded allocation. When that happens, we request a new allocation to the GarbageCollector (which forces GC to run if needed).
3. We keep doing this until we finish downloading the file. The larger the file, the more opportunities we'll give the GC to manage the space.
4. Once we have finished downloading the file, we store the final size and persist it to the state of the dagstore.
5. The next time we download this mount, we can ask the GC to allocate/reserve the exact size.
6. If the mount errors, we delete the stored size, as this could be an indication that the downloaded file was corrupt.

aschmahmann:
@raulk I think that seems reasonable although IIUC there's an edge case where I start downloading 100 CARs at once and while none of them are completed I'm also just out of space. In this case you'll probably need to do something like start moving some of the transients into a "hung" state and clearing out their space so the other CARs can mount

raulk:
@aschmahmann That edge case should be partially covered by the `MaxConcurrentReadyFetches` configuration attribute which limits the number of concurrent mount fetches. However, the GC should track inflight reservations and be intelligent enough to know when it needs to reject new requests (e.g. imagine I'm downloading 6 shards 1TiB large each, with a quota of 3TiB; the clever outcome is to eventually allow 3 shards to be fully downloaded).

aarshkshah1992:
Subsumed by https://github.com/filecoin-project/dagstore/pull/124.

PR #135, opened 2022-08-08
fix: return result from successful destroy

LexLuthr:
Currently, even though we provide a channel to monitor for DestroyShard(), the result is only returned if there is an error.
In case, we are monitoring the channel for successful deletion, then no value it returned to the channel and goroutine stays waiting.


aarshkshah1992 (approved):

jacobheun:
Closing in favor of https://github.com/filecoin-project/dagstore/pull/136

PR #145, opened 2022-09-19
Fix: add on-disk deletion of shards

LexLuthr:
Currently, dagstore implementation only deletes the shard from in-memory map of keys. When dagstore is restarted, the shard comes back from disk.

### DIFF
--- a/dagstore_control.go
+++ b/dagstore_control.go
@@ -300,10 +300,9 @@ func (d *DAGStore) control() {
 			d.lk.Lock()
 			delete(d.shards, s.key)
 
-			// Delete the entry directly from the datastore as persist does not implement deletion
-			// Persist should not make any changes for the control_loop runs for deletion
+			// Perform on-disk delete after the switch statement. This is only in-memory delete.
 			d.lk.Unlock()
-			res := &ShardResult{Key: s.key, Error: err}
+			res := &ShardResult{Key: s.key, Error: nil}
 			d.dispatchResult(res, tsk.waiter)
 			// TODO are we guaranteed that there are no queued items for this shard?
 
@@ -312,8 +311,7 @@ func (d *DAGStore) control() {
 
 		}
 
-		// persist the current shard state. Skip if op is OpShardDestroy. Otherwise, it
-		// re-registers the shard
+		// persist the current shard state. If Op is OpShardDestroy then delete directly from DB.
 		if tsk.op == OpShardDestroy {
 			if err := d.store.Delete(d.ctx, datastore.NewKey(s.key.String())); err != nil {
 				log.Errorw("DestroyShard: failed to delete shard from database", "shard", s.key, "error", err)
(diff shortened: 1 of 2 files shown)


dagstore_control.go
```diff
@@ -297,8 +299,12 @@ func (d *DAGStore) control() {
 
 			d.lk.Lock()
 			delete(d.shards, s.key)
+
+			// Delete the entry directly from the datastore as persist does not implement deletion
+			// Persist should not make any changes for the control_loop runs for deletion
+			err = d.store.Delete(d.ctx, datastore.NewKey(s.key.String()))
```
dirkmc, line 305: Currently the dagstore assumes that operations against the datastore will succeed, and just logs an error if they fail.
So I'd suggest we continue to follow that pattern here. Let's move the call to `s.store.Delete` to the if / else block on line 318

dirkmc (commented):
Looks great, just one suggestion for a refactor. Test looks great too 👍

dagstore_control.go
```diff
@@ -297,8 +299,11 @@ func (d *DAGStore) control() {
 
 			d.lk.Lock()
 			delete(d.shards, s.key)
+
+			// Delete the entry directly from the datastore as persist does not implement deletion
+			// Persist should not make any changes for the control_loop runs for deletion
 			d.lk.Unlock()
-			res := &ShardResult{Key: s.key, Error: nil}
+			res := &ShardResult{Key: s.key, Error: err}
```
LexLuthr, line 306:
```suggestion
			res := &ShardResult{Key: s.key, Error: nil}
```

dagstore_control.go
```diff
@@ -307,9 +312,16 @@ func (d *DAGStore) control() {
 
 		}
 
-		// persist the current shard state.
-		if err := s.persist(d.ctx, d.config.Datastore); err != nil { // TODO maybe fail shard?
-			log.Warnw("failed to persist shard", "shard", s.key, "error", err)
+		// persist the current shard state. Skip if op is OpShardDestroy. Otherwise, it
```
LexLuthr, line 315:
```suggestion
		// persist the current shard state. If Op is OpShardDestroy then delete directly from DB.
```

dagstore_control.go
```diff
@@ -307,9 +312,16 @@ func (d *DAGStore) control() {
...
+		// re-registers the shard
```
LexLuthr, line 316:
```suggestion
```

dagstore_control.go
```diff
@@ -297,8 +299,11 @@ func (d *DAGStore) control() {
 
 			d.lk.Lock()
 			delete(d.shards, s.key)
+
+			// Delete the entry directly from the datastore as persist does not implement deletion
```
LexLuthr, line 303:
```suggestion
```

dagstore_control.go
```diff
@@ -297,8 +299,11 @@ func (d *DAGStore) control() {
...
+			// Persist should not make any changes for the control_loop runs for deletion
```
LexLuthr, line 304:
```suggestion
			// Perform on-disk delete after the switch statement. This is only in-memory delete.
```

dirkmc (approved):

PR #150, opened 2023-02-08
feat: add expiration support to index backed blockstore

jacobheun:
This is a breaking change to the IBB interface.

This pulls the LRU interface from github.com/hnlq715/golang-lru but includes a constructor for both expiry and an evict callback.


dirkmc (approved):
Looks great! 👍

PR #153, opened 2023-03-07
change index-backed blockstore dagstore interface

dirkmc:
Change ShardsContainingHash to ShardsContainingCid so that we can detect identity cids


jacobheun (approved):
Note for versioning, this is a breaking api change.

PR #155, opened 2023-04-03
feat: `io.Reader` factory for `ShardAccessor`

distractedm1nd:
At celestia-node, we don't only use `ShardAccessor`'s Blockstore to serve individual blocks, but also use the underlying mount to stream the CAR file to other full nodes. We do this by adding a `Reader() io.Reader` method to `ShardAccessor` that creates independent mmap-backed (when possible) readers. We would love to upstream it to avoid maintaining a fork.

### DIFF
--- a/accessor.go
+++ b/accessor.go
@@ -2,6 +2,7 @@ package dagstore
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"sync"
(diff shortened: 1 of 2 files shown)


accessor.go
```diff
@@ -54,11 +54,34 @@ func (sa *ShardAccessor) Shard() shard.Key {
...
+
+	if sa.mmapr == nil {
+		if f, ok := sa.data.(*os.File); ok {
+			if mmapr, err := mmap.Open(f.Name()); err != nil {
+				log.Warnf("failed to mmap reader of type %T: %s; using reader as-is", sa.data, err)
+			} else {
+				sa.mmapr = mmapr
+			}
+		}
+	}
```
raulk, line 69: Let's extract this logic to a `tryMmap()` method (and document it requires the lock to be held) and refactor the `Blockstore()` method to use that?

accessor.go
```diff
@@ -54,11 +54,34 @@ func (sa *ShardAccessor) Shard() shard.Key {
...
+
+	if sa.mmapr != nil {
+		return sa.mmapr.ReadAt(p, 0)
+	}
```
raulk, line 73: Would move this above since we expect it to be the most common path when a shard gets reused.

accessor.go
```diff
@@ -54,11 +54,34 @@ func (sa *ShardAccessor) Shard() shard.Key {
...
+}
+
 func (sa *ShardAccessor) Blockstore() (ReadBlockstore, error) {
 	var r io.ReaderAt = sa.data
 
 	sa.lk.Lock()
-	if f, ok := sa.data.(*os.File); ok {
+	if sa.mmapr != nil {
+		r = sa.mmapr
+	} else if f, ok := sa.data.(*os.File); ok {
```
raulk, line 84: Let's extract the mmap logic; then make this method short-circuit for more readable and idiomatic code.

Wondertan (approved):
LGTM now

PR #156, opened 2023-04-07
Allow configuration of a custom shard indexer

willscott:
Supersedes #154

Allows setting a custom function for the dagstore to use to index shards


dirkmc (approved):
LGTM 👍

PR #157, opened 2023-04-13
fix: `FileMount` should use url `Path` instead of `Host`

distractedm1nd:
We [encountered a bug](https://github.com/celestiaorg/celestia-node/issues/2031) recently after switching from `FSMount` to `FileMount`.

The root cause of the problem was the usage of the `Host` field of `url.URL` to store the file path. This led to incorrect URL encoding, which caused the application to fail when trying to recover the state of shards. 

Should we add backwards compatibility to the `Deserialize` method so that mounts previously `Serialized()` using `Host` will be loaded correctly?

### DIFF
--- a/mount/file.go
+++ b/mount/file.go
@@ -45,7 +45,7 @@ func (f *FileMount) Serialize() *url.URL {
 
 func (f *FileMount) Deserialize(u *url.URL) error {
 	if u.Path == "" {
-		return fmt.Errorf("invalid host")
+		return fmt.Errorf("invalid path")
 	}
 	f.Path = u.Path
 	return nil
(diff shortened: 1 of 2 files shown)


raulk:
Great find! The most correct way to handle the change would be to perform a shard store migration. @nonsense @dirkmc I don't supposed we've encountered this issue in Boost or lotus-miner since we use a Filecoin-specific mount type?

dirkmc:
Yes, as you say we haven't come across this issue because we use a specific mount type

mount/file.go
```diff
@@ -39,15 +39,15 @@ func (f *FileMount) Stat(_ context.Context) (Stat, error) {
...
 	return &url.URL{
-		Host: f.Path,
+		Path: f.Path,
 	}
 }
 
 func (f *FileMount) Deserialize(u *url.URL) error {
-	if u.Host == "" {
+	if u.Path == "" {
 		return fmt.Errorf("invalid host")
```
Wondertan: Update error?

distractedm1nd: fixed

